PR #1, opened 2019-05-06
Print error string in die() function

siteshwar:


kdudka:
Good catch.  Thanks for the contribution!

PR #22, opened 2021-07-20
Add support for custom filters in JSON files

lzaoral:
This is just a draft to see if there are no major issues. I'll add tests if everything will look alright.

Format example:
```json
{
    "msg-filter" : [
        {
            "checker" : "DIVINE|SYMBIOTIC",
            "regexp" : "memory"
        },
        {
            "checker" : "COMPILER_WARNING",
            "regexp" : "called on unallocated object",
            "replace" : "called correctly, no UB here"
        }
    ]
}
```


kdudka:
Nice!  My only suggestion is to make `MsgFilter::setJSONFilter()` catch `boost::regex_error` to provide a more user-friendly error message when the given regex fails to compile.  Instead of:
```sh
$ csgrep -f <(echo '{"msg-filter":[{"checker":"","regexp":"*"}]}')
terminate called after throwing an instance of 'boost::wrapexcept<boost::regex_error>'
  what():  The repeat operator "*" cannot start a regular expression.  The error occurred while parsing the regular expression: '>>>HERE>>>*'.
zsh: IOT instruction (core dumped)  csgrep -f <(echo '{"msg-filter":[{"checker":"","regexp":"*"}]}')

$ echo $?
134
```
... I would expect something like this:
```sh
$ csgrep -f <(echo '{"msg-filter":[{"checker":"","regexp":"*"}]}')
filter.js: error: Failed to compile regex: The repeat operator "*" cannot start a regular expression.  The error occurred while parsing the regular expression: '>>>HERE>>>*

$ echo $?
1
```

@praiskup What is your take on this?

src/csdiff.cc
```diff
@@ -151,6 +153,13 @@ int main(int argc, char *argv[])
     const bool showInternal = vm.count("show-internal");
     const bool silent       = vm.count("quiet");
 
+    if (vm.count("filter-file")) {
```
praiskup, line 156: Can this be specified multiple times?

praiskup: Ah, it can. Nice.

praiskup:
This is nice, thank you @lzaoral for working on this!

I'd suggest writing some test case for this, and perhaps document the format somewhere so it is easier to use.
Though I think I can personally adapt ;-)

kdudka:
@praiskup Thanks for review!  Covering this by tests is already planned as it is written above.  And yes, the csdiff project has always been pretty bad regarding documentation of the supported data formats.  Any improvements in this regard would certainly be appreciated.

lzaoral:
Thanks for the reviews @kdudka and @praiskup! I've rebased the PR to include tests, fix the `boost::regex_error` issue and to document the expected JSON format in respective manual pages.

kdudka:
Amazing progress.  Thank you for working on this!

My only concern is that `"checker" : "DIVINE|SYMBIOTIC"` currently does not work as expected.  It matches the string literally rather than as a regex.  This is not your fault.  I forgot to say that this is something that yet needs to be implemented.  Could you please check pull request #23 which implements it?

If you are fine with it, I would merge #23 first and then rebase this pull request on top of it.  There are no conflicts or (test breakage) between these pull requests.

We could probably extend the `--filter-file` tests to cover the regex-based matching of checkers, too.

lzaoral:
I've rebased the PR to include #22  and #23. Also previously, I forgot to add the new `csgrep` tests to the corresponding `CMakeLists.txt`.

praiskup:
This should work for us, thank you @lzaoral!

lzaoral:
I've just changed the documentation to include meaning of the missing `replace` entry in the JSON format.

kdudka:
Merging, thanks for the contribution!

PR #27, opened 2021-07-29
travis: run copr build only for 'master' branch

praiskup:
travis: run copr build only for the master branch
    
    This is needed to avoid building into the @codescan/csdiff copr project
    upon push to non-master branch.  Also, pull-requests against non-master
    branch would be excluded, but they should be rare.
    
    Drop the -x flag from the Travis build.  That PR webhook isn't a private
    information but still..  it is better to not make a honeypot from all
    our Travis logs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ matrix:
           dist: focal
           env:
               - COPR_BUILD=yes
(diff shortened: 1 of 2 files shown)


kdudka:
Merging, thanks!

PR #30, opened 2021-09-30
travis: don't require GitHub webhook payload

praiskup:
This is not needed when the build is triggered manually in Copr.


kdudka:
Merging, thanks!

PR #31, opened 2021-12-06
tests: Use json file extension for stored scan-results

jamacku:
It would be better to use JSON file extension (``.json``) rather then ``.js``.

It would allow Github to show more accurate statistics of csdiff project.
Also it would allow IDE to better parse those files, since json object without assignment isn't valid syntax in ``.js`` files. It is understand as code block.

> Note: I didn't test these changes


jamacku:
[CI result](https://download.copr.fedorainfracloud.org/results/@codescan/csdiff-pull-requests/srpm-builds/03007051/builder-live.log.gz)
```
...
The following tests FAILED:
	275 - cssort-5.8-03-by-checker (Failed)
	276 - cssort-5.8-03-by-path (Failed)
...
```

kdudka:
We cannot easily change this in csmock because `scan-results.js` is the documented interface of it.  But changing this in csdiff's test-suite should not break anything and if it helps users, I am all for it!  Merging it with a minor fixup.  Thanks for the contribution!

jamacku:
Great, thank you for clarifying.

Issue #37, opened 2022-03-29
Unknown event not supported


psimovec:
```
❯ csgrep pid-3507314.out.conv
pid-3507314.out.conv:2: parse error: wrong token: T_UNKNOWN (expected T_EVENT)
Error: SYMBIOTIC_WARNING:
/builddir/build/BUILD/gnupg-2.3.2/common/membuf.c:114: note: call stack: function put_membuf (=0, =0, =3)
/builddir/build/BUILD/gnupg-2.3.2/common/membuf.c:136: note: call stack: function put_membuf_str (=0, =0)
/builddir/build/BUILD/gnupg-2.3.2/common/mapstrings.c:156: note: call stack: function map_static_macro_string (=0)
/builddir/build/BUILD/gnupg-2.3.2/common/t-mapstrings.c:68: note: call stack: function test_map_static_macro_string ()
/builddir/build/BUILD/gnupg-2.3.2/common/t-mapstrings.c:97: note: call stack: function main (=2, =0)
<Unknown>:<Unknown>: note: Additional Info: address: 255:0
<Unknown>:<Unknown>: note: Additional Info: pointing to: none
/builddir/build/BUILD/gnupg-2.3.2/common/membuf.c:52:13: note: Non-deterministic values: gcry_malloc: len 8 bytes, [0xff|7 times 0x0] (i64: 255)
/builddir/build/BUILD/gnupg-2.3.2/common/membuf.c:52:13 (offset): note: Non-deterministic values: gcry_malloc: len 8 bytes, [8 times 0x0] (i64: 0)
```

```
Error: SYMBIOTIC_WARNING:
<Unknown>:<Unknown>: error: memory error: out of bound pointer
/builddir/build/BUILD/gnupg-2.3.2/common/membuf.c:114: note: call stack: function put_membuf (=0, =0, =3)
/builddir/build/BUILD/gnupg-2.3.2/common/membuf.c:136: note: call stack: function put_membuf_str (=0, =0)
/builddir/build/BUILD/gnupg-2.3.2/common/mapstrings.c:156: note: call stack: function map_static_macro_string (=0)
/builddir/build/BUILD/gnupg-2.3.2/common/t-mapstrings.c:68: note: call stack: function test_map_static_macro_string ()
/builddir/build/BUILD/gnupg-2.3.2/common/t-mapstrings.c:97: note: call stack: function main (=2, =0)
<Unknown>:<Unknown>: note: Additional Info: address: 255:0
[...]

PR #38, opened 2022-03-29
csparser: do not reject <Unknown> produced by symbiotic2cs

kdudka:
Fixes: https://github.com/csutils/csdiff/issues/37
Closes: https://github.com/csutils/csdiff/pull/38


lzaoral:
I suggest dropping this PR in favour of https://github.com/aufover/rpm-symbiotic/pull/6 so that the format stays consistent across all of our output converters.

kdudka:
The proposed change is a no-op if the input is correct.  On the other hand, I see no benefit in rejecting `<Unknown>`.  It would only complicate processing of results obtained by an old version of `symbiotc2cs`.  So I would prefer to merge both the pull requests.

lzaoral:
Alright. Merging,

kdudka:
Thanks!

PR #40, opened 2022-04-13
Port the project to C++11 - Part 1

lzaoral:
This PR ports `src/abstract-*` to C++ 11.

### DIFF
--- a/src/abstract-filter.cc
+++ b/src/abstract-filter.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2011 Red Hat, Inc.
+ * Copyright (C) 2011-2022 Red Hat, Inc.
  *
  * This file is part of csdiff.
  *
@@ -174,7 +174,7 @@ void CtxEmbedder::handleDef(const Defect &defOrig)
         return;
     }
 
-    std::fstream fstr(evt.fileName.c_str(), std::ios::in);
+    std::ifstream fstr(evt.fileName);
     if (!fstr) {
         // failed to open input file
         agent_->handleDef(defOrig);
--- a/src/abstract-filter.cc
+++ b/src/abstract-filter.cc
@@ -35,11 +35,10 @@ struct PredicateFilter::Private {
     using IPredicatePtr = std::unique_ptr<IPredicate>;
     using TList         = std::vector<IPredicatePtr>;
 
-    bool                invertEach_;
+    bool                invertEach_ = false;
     TList               preds_;
 
-    Private():
-        invertEach_(false)
+    Private()
     {
     }
 };
--- a/src/abstract-filter.cc
+++ b/src/abstract-filter.cc
@@ -32,7 +32,9 @@
 // implementation of PredicateFilter
 
 struct PredicateFilter::Private {
-    typedef std::vector<IPredicate *>               TList;
+    using IPredicatePtr = std::unique_ptr<IPredicate>;
+    using TList         = std::vector<IPredicatePtr>;
+
     bool                invertEach_;
     TList               preds_;
 
@@ -48,17 +50,11 @@ PredicateFilter::PredicateFilter(AbstractWriter *agent):
 {
 }
 
-PredicateFilter::~PredicateFilter()
-{
-    for (IPredicate *pred : d->preds_)
-        delete pred;
-
-    delete d;
-}
+PredicateFilter::~PredicateFilter() = default;
 
 void PredicateFilter::append(IPredicate *pred)
 {
-    d->preds_.push_back(pred);
+    d->preds_.emplace_back(pred);
 }
 
 void PredicateFilter::setInvertEachMatch(bool enabled)
@@ -70,7 +66,7 @@ bool PredicateFilter::matchDef(const Defect &def)
 {
     const bool neg = d->invertEach_;
 
-    for (const IPredicate *pred : d->preds_) {
+    for (const auto &pred : d->preds_) {
         if (neg == pred->matchDef(def))
             return false;
     }
--- a/src/abstract-filter.cc
+++ b/src/abstract-filter.cc
@@ -38,9 +38,7 @@ struct PredicateFilter::Private {
     bool                invertEach_ = false;
     TList               preds_;
 
-    Private()
-    {
-    }
+    Private() = default;
 };
 
 PredicateFilter::PredicateFilter(AbstractWriter *agent):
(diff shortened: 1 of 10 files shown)


src/abstract-parser.hh
```diff
@@ -55,10 +55,10 @@ class AbstractParser {
         }
 
     protected:
-        AbstractParser() { }
+        AbstractParser() = default;
 
     private:
-        const TScanProps emptyProps_;
+        const TScanProps emptyProps_{};
```
kdudka, line 61: Just curious, why do we need the extra brackets?

lzaoral: Defaulted constructors do not initialize non-`static` and `const` members. From https://en.cppreference.com/w/cpp/language/default_constructor:

>The implicitly-declared or defaulted (since C++11) default constructor for class T is defined as deleted if any of the following is true:
>
>    T has a non-[const-default-constructible](https://en.cppreference.com/w/cpp/language/default_initialization#Default_initialization_of_a_const_object) const member without a default member initializer (since C++11).

TL;DR non-`static` but `const` members of types with non-user provided default constructors won't be initialized by the defaulted default constructor so you have to default-initialize them explicitly by default member initializer.

Yeah, C++ is a nice language. 🙃

src/abstract-filter.cc
```diff
@@ -186,7 +186,6 @@ void CtxEmbedder::handleDef(const Defect &defOrig)
     dropCtxLines(&def.events);
     appendCtxLines(&def.events, fstr, evt.line, ctxLines_ - 1);
 
-    // close the file stream and forward the result
-    fstr.close();
```
kdudka, line 190: I do not think the call to `fstr.close()` is redundant.  If you remove it, the file will be unnecessarily kept open until `agent_->handleDef(def)` finishes.  There can be a big chain of decorators behind the call.

src/abstract-tree.hh
```diff
@@ -35,12 +35,8 @@ class AbstractTreeDecoder {
...
         /// read scan properties if available
         virtual void readScanProps(
-                TScanProps             *pDst,
-                const pt::ptree        *root)
-        {
-            (void) pDst;
-            (void) root;
-        }
+                TScanProps      * /* pDst */,
+                const pt::ptree * /* root */) { }
```
kdudka, line 39: I would remove this change.  The previous syntax feels more readable to me and did not cause any problems, as far as I know.

src/abstract-filter.hh
```diff
@@ -57,7 +57,7 @@ class GenericAbstractFilter: public AbstractWriter {
 /// decorator
 class EventPrunner: public GenericAbstractFilter {
     private:
-        int thr_;
+        unsigned thr_;
```
kdudka, line 60: This change triggers a new warning:
```
src/abstract-filter.cc: In member function ‘virtual void EventPrunner::handleDef(const Defect&)’:
src/abstract-filter.cc:87:32: warning: comparison of integer expressions of different signedness: ‘const int’ and ‘unsigned int’ [-Wsign-compare]
   87 |         if (evt.verbosityLevel <= thr_)
      |             ~~~~~~~~~~~~~~~~~~~^~~~~~~
```

kdudka (changes requested):
Thank you for working on this!  Overall it looks good.  A few minor changes suggested inline.

src/abstract-filter.hh
kdudka: I think the proper fix would be this:
```diff
--- a/src/abstract-filter.hh
+++ b/src/abstract-filter.hh
@@ -61,7 +61,7 @@ class EventPrunner: public GenericAbstractFilter {
         int thr_;

     public:
-        EventPrunner(AbstractWriter *agent, unsigned thr):
+        EventPrunner(AbstractWriter *agent, int thr):
             GenericAbstractFilter(agent),
             thr_(thr)
         {
```

src/abstract-filter.cc
lzaoral: I did not take into account this possibility. I'll revert this change.

src/abstract-tree.hh
lzaoral: Yes, it's only a cosmetic change but I think it makes the code more concise.

src/abstract-filter.cc
lzaoral: Change dropped.

src/abstract-tree.hh
lzaoral: Commit dropped,

src/abstract-filter.hh
lzaoral: Addressed in 1fd0a96.

src/csparser.hh
```diff
@@ -30,6 +30,10 @@ class CovParser: public AbstractParser {
         bool getNext(Defect *);
         bool hasError() const;
 
+        virtual EFileFormat inputFormat() const {
```
kdudka, line 33: Should we spell this as `EFileFormat inputFormat() const override {` to be consistent with commit ecd971c4b951ed5b54db5ad5648b6f9f7d9c53ee?

src/gcc-parser.hh
```diff
@@ -30,6 +30,10 @@ class GccParser: public AbstractParser {
         bool getNext(Defect *);
         bool hasError() const;
 
+        virtual EFileFormat inputFormat() const {
```
kdudka, line 33: ditto

src/json-parser.hh
```diff
@@ -31,6 +31,10 @@ class JsonParser: public AbstractParser {
         virtual bool hasError() const;
         virtual const TScanProps& getScanProps() const;
 
+        virtual EFileFormat inputFormat() const {
```
kdudka, line 34: ditto

kdudka (approved):

src/csparser.hh
lzaoral: Definitely! I want to make this change as well but in a separate PR as this one is too big already (and focuses mainly on `src/abstrac-*` related changes). Is that ok with you?

kdudka: This is new code you are introducing in this pull request.  I would prefer use the desired form since the beginning, rather than adding new code that we already plan to rework in a future pull request.  If it depends on other changes, the commit in question can be moved to the next pull request, to avoid unnecessary changes back and forth.

lzaoral: Ok, let's drop this for now.

lzaoral:
Thanks for the reviews, @kdudka!

kdudka:
No problem.  Thank you for modernizing the code!

Issue #42, opened 2022-04-19
JSON writer produces output that JSON parser fails to read


kdudka:
This happens mainly with `csgrep --embed-context ...` when it reads some garbage from source files (e.g. from `qt3d-everywhere-src-5.15.3/src/3rdparty/assimp/code/FBXMaterial.cpp:326` from `qt5-qt3d-5.15.3-1.el8`).  We should fix `--embed-context` to filter out the garbage, as well as make the JSON reader accept whatever the JSON writer produces.

Minimal example:
``` sh
$ echo $'\xb4' > x.c && echo 'x.c:1: warning: msg' | csgrep --mode=json -U1 | csgrep
-:19: error: invalid code sequence
```

PR #46, opened 2022-04-21
json-writer: replace invalid sequences in messages with �

lzaoral:
Fixes: #42


kdudka (changes requested):
I like the solution but it needs to be extended to `sarifEncodeMsg()`, too:
```sh
$ echo $'\xb4' > x.c && echo 'x.c:1: warning: msg' | csgrep --mode=sarif -U1 | csgrep
-:71: error: invalid code sequence
```

src/json-writer.cc
```diff
@@ -28,8 +28,18 @@
...
 #include <boost/iostreams/filter/regex.hpp>
+#include <boost/nowide/utf/convert.hpp>
 #include <boost/property_tree/json_parser.hpp>
 
+static inline std::string sanitizeUTF8(const std::string &str)
+{
+    using boost::nowide::utf::convert_string;
+
+    // every non-UTF8 sequence will be replaced with �
+    // � - REPLACEMENT CHARACTER U+FFFD
```
kdudka, line 39: I would prefer not to have the special character directly in the source code.  Could we just write `0xEF 0xBF 0xBD` in the comment?

kdudka:
I am marking this as draft until we resolve the compatibility with EPEL buildroots.

src/json-writer.cc
lzaoral: Addressed in 620dc470d46908c64d3a5d14c78e015c71836aee.

kdudka:
Thanks!  I hit this difference in behavior while I was working on PR #48 yesterday.

lzaoral:
Thanks for the review, @kdudka! Merging.

lzaoral:
Thanks @praiskup! Addressed in f7e5c14fbbd0c0f48afdba50e4e7d427da48fce0. I'll merge this PR when it's green.

PR #49, opened 2022-05-04
track a separate "tool" attribute of defects [WIP]

kdudka:
It can be propagated by csdiff's native JSON format and statically
assigned or guessed for the other formats.


lzaoral (approved):
LGTM! Thanks, @kdudka!

PR #52, opened 2022-06-01
cwe-mapper: do not assign unrelated CWEs for GCC and ShellCheck

kdudka:
Due to historical reasons, we distinguish checkers (a.k.a. rules) by the checker field for Coverity only.  For other tools we map this to the key events, such as `warning[-Wshadow]` in case of gcc.

The CWE mapping is defined for checker/keyEvent pairs.  For Coverity, it is fine to assign CWE based on checker only in case the exact pair is not matched.  For other tools, it does not make any sense, because the checker field maps to the tool and the actual checkers (rules) can be distinguished by the key event only.

This is a follow-up to commit 142ee3f8a5f31dcc3f93f68122ae075d79a5fd39, which landed when we had CWE mapping for Coverity and Cppcheck only.

Reported-by: Steve Grubb


lzaoral (approved):
LGTM! Thanks, @kdudka!

kdudka:
Thanks for review!  Merging...

PR #53, opened 2022-06-02
data/cwe-names.csv: sync with coverity-analytics

kdudka:
Contributed-by: Steve Grubb


kdudka:
@stevegrubb As coverity-analytics depends on csdiff anyway, we might consider maintaining the data in this repository.  All systems with csdiff installed already provide the file as `/usr/share/csdiff/cwe-names.csv`.

PR #55, opened 2022-06-02
Replace current Fedora Copr integration integration with Packit

lzaoral:
Packit seems to be more stable and reliable at the moment.


lzaoral:
Thank you, @kdudka! Merging...

.github/workflows/copr-build.yml
```diff
@@ -1,24 +0,0 @@
...
-    runs-on: ubuntu-latest
-
-    steps:
-      - name: Check out proper version of sources
-        uses: actions/checkout@v3
-
-      - name: Submit the build
-        env:
-          COPR_PR_WEBHOOK: https://copr.fedorainfracloud.org/webhooks/custom/46962/a72fe511-c141-4f27-b7cf-de03c065e64e/csdiff/
-          COPR_PUSH_WEBHOOK: ${{ secrets.COPR_PUSH_WEBHOOK }}
```
praiskup, line 21: This has been replaced with no alternative.  https://copr.fedorainfracloud.org/coprs/g/codescan/csdiff/ stays orphaned.  I'm going to try to fix this.

praiskup:
But indeed, this is a good step forward!  Much simpler and modern configuration.

.github/workflows/copr-build.yml
kdudka: As I understand it, Packit creates temporary COPR projects:
- for the `main` branch: https://copr.fedorainfracloud.org/coprs/packit/csutils-csdiff-main/
- and for each pull request, as for example: https://copr.fedorainfracloud.org/coprs/packit/csutils-csdiff-56/

So it should more or less cover the uses cases that we initially had https://copr.fedorainfracloud.org/coprs/g/codescan/csdiff/ for.

kdudka: OK, I can see pull request #56 now.  So the idea is to use our own COPR namespace again, rather than the Packit's namespace.

lzaoral: Thank you @praiskup! @kdudka, on the other hand, we could use the `@codescan` namespace as a **permanent** place for latest commits from the `main` branches of all projects in the `csutils` organization.

praiskup: The idea was to make the "trusted" builds from merged commits "easily consumable" (I personally always run the pre-release csdiff locally, by `sudo dnf copr enable @codescan/csdiff`).

kdudka: That sounds useful indeed.  Thank you for working on it!

PR #56, opened 2022-06-14
CI: build committed packages into @codescan/csdiff

praiskup:

PR #57, opened 2022-06-14
CI: make Packit use the same `%{version}.%{release}` as `make-srpm.sh`

lzaoral:
We should be using the same versioning scheme so that we don't unintentionally break package upgrades.

Blocks: #56

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -13,8 +13,10 @@ upstream_package_name: csdiff
 # downstream (Fedora) RPM package name
 downstream_package_name: csdiff
 
+release_suffix: ""
 actions:
     post-upstream-clone: ./make-srpm.sh --generate-spec
+    get-current-version: "sed -n 's|^Version: *||p' csdiff.spec"
 
 jobs:
     - &copr


.packit.yaml
```diff
@@ -13,8 +13,10 @@ upstream_package_name: csdiff
 # downstream (Fedora) RPM package name
 downstream_package_name: csdiff
 
+release_suffix: ""
 actions:
     post-upstream-clone: ./make-srpm.sh --generate-spec
+    get-current-version: sh -c "grep 'Version:' csdiff.spec | cut -d':' -f2 | tr -d ' '"
```
kdudka, line 19: Could we simplify the command as follows?
```sh
sed -n 's|^Version: *||p' csdiff.spec
```

praiskup:
Nice, thank you. Example output `Version: 2.5.0.20220614.180034.ge02ff8a.pr_57`.

.packit.yaml
lzaoral: Thanks! I didn't know this syntax of `sed`. Addressed in 36e5245. Note that the whole command has to be passed as a string because `'Version: '` was being interpreted as invalid YAML:
```
yaml.scanner.ScannerError: mapping values are not allowed here
  in "<unicode string>", line 19, column 44:
     ... rent-version: sed -n "s|^Version: *||p" csdiff.spec
                                         ^

During handling of the above exception, another exception occurred:
```

kdudka (approved):
Looks good to me.  Thanks!

PR #58, opened 2022-06-15
packit: metadata section is no longer needed

TomasTomecek:
We recently dropped the need for "metadata", now all configuration can happen on the job level.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -22,20 +22,18 @@ jobs:
     - &copr
       job: copr_build
       trigger: pull_request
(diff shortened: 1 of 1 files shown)


TomasTomecek:
And thanks to @praiskup, this also fixes the chroot set for the builds triggered by commits to main. We just discussed on #packit at libera.

lzaoral:
Thank you, @TomasTomecek! Accidentally, I've just pushed two commits to `main` with the exact same changes so I'm going to close this PR.

PR #59, opened 2022-06-16
doc: use C locale when generating manual pages

lzaoral:
Otherwise, e.g. month could be in other language when `LC_TIME` was set to such locale.

Fixes: `Wide character in print at /usr/bin/help2man line 672.`


kdudka (approved):
Looks good.  Thanks!

PR #60, opened 2022-06-23
make-srpm.sh: `boost-python3-devel` was merged into `boost-devel` in Fedora 33

lzaoral:
... and the corresponding obsoletes tag was removed in Fedora 37.


kdudka (approved):
Looks good.  Thanks!

PR #61, opened 2022-06-29
Add some missing name mappings

stevegrubb:
During a recent scan, I found several missing name mappings.
This adds 12 or so mappings.


data/cwe-names.csv
```diff
@@ -468,10 +479,12 @@
...
 "1220","Insufficient Granularity of Access Control"
+"1228","API / Function Errors"
 "1230","Exposure of Sensitive Information Through Metadata"
 "1235","Incorrect Use of Autoboxing and Unboxing for Performance Critical Operations"
 "1236","Improper Neutralization of Formula Elements in a CSV File"
 "1240","Use of a Risky Cryptographic Primitive"
 "1241","Use of Predictable Algorithm in Random Number Generator"
 "1265","Unintended Reentrant Invocation of Non-reentrant Code Via Nested Calls"
 "9001","Low Level Non-security Compiler Warning"
+"unmapped","unknown"
```
kdudka, line 490: How did this line appear here?  It makes csdiff print an error upon loading the table:
```
data/cwe-names.csv:490: error: invalid CWE ID
```

stevegrubb: We have a catchall that puts unmapped into any NA. They in turn do not map to any name and we get another NA.

stevegrubb: This usually comes from a group of ancient gcc warnings that do not have any -Wno  option associated with them. Because they are free-form text, they do not map to any CWE.

kdudka:
Merging, thanks!

PR #62, opened 2022-07-08
json-parser: decode the JSON format produced by GCC

kdudka:


src/json-parser.cc
```diff
@@ -93,6 +93,12 @@ class SarifTreeDecoder: public AbstractTreeDecoder {
         ImpliedAttrDigger           digger;
 };
 
+/// tree decoder of the JSON format produced by GCC
+class GccTreeDecoder: public AbstractTreeDecoder {
+    public:
+        virtual bool readNode(Defect *def, pt::ptree::const_iterator defIter);
```
lzaoral, line 99:
```suggestion
        bool readNode(Defect *def, pt::ptree::const_iterator defIter) override;
```

src/json-parser.cc
```diff
@@ -638,3 +653,66 @@ bool SarifTreeDecoder::readNode(
 
     return true;
 }
+
+bool gccReadEvent(DefEvent *pEvt, const pt::ptree &evtNode)
```
lzaoral, line 657:
```suggestion
static bool gccReadEvent(DefEvent *pEvt, const pt::ptree &evtNode)
```

tests/csgrep/94-gcc-json-stdin.txt
```diff
@@ -0,0 +1,250 @@
...
+          "file": "xxx.c",
+          "column": 11
+        }
+      }
+    ],
+    "column-origin": 1,
+    "option": "-Wunused-parameter",
+    "children": [],
+    "option_url": "https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wunused-parameter",
+    "message": "unused parameter \u2018argc\u2019"
```
lzaoral, line 87: OT: Hmm, `\u2018argc\u2019` doesn't look right to me.  Looks like a bug in GCC's output. (However, `csgrep` still converts the bytes to actual characters which is nice!)

src/json-parser.cc
lzaoral, line 99:
```suggestion
        bool readNode(Defect *def, pt::ptree::const_iterator defIter) override;
```

lzaoral (changes requested):
Thank you, @kdudka!  Please update the copyright year in `src/json-parser.cc`. Additional minor nitpicks inline.

tests/csgrep/94-gcc-json-stdin.txt
kdudka: Those escape sequences are not produced by GCC but Python's `json` module, which I used to reformat the input to be human-readable:
```py
import json
o = json.loads(open("xxx.js").read())
print(json.dumps(o, indent=2))
```

kdudka:
@lzaoral Thanks for review!  I have updated the pull requests per you comments.

lzaoral (approved):

PR #63, opened 2022-07-11
json-parser: distinguish warnings from GCC compiler and analyzer

kdudka:
... in the JSON format produced by GCC


kdudka:
Thanks for review!  Merging...

PR #65, opened 2022-07-28
Modernize CMake

lzaoral:


src/CMakeLists.txt
```diff
@@ -188,8 +189,10 @@ macro(create_manpage BINARY)
             > ${BINARY}.1 || rm -f ${BINARY}.1
         COMMENT "Generating ${BINARY} man page"
         VERBATIM)
+
+    include(GNUInstallDirs)
```
kdudka, line 180: There is already `include(GNUInstallDirs)` before `install(...)`.

src/CMakeLists.txt
```diff
@@ -148,31 +128,20 @@ failed to query PYTHON_SITEARCH")
 
     message(STATUS "BOOST_PYTHON: ${BOOST_PYTHON}")
 
-    # verify PYTHON_LIBRARY
-    if ("${PYTHON_LIBRARY}" STREQUAL "PYTHON_LIBRARY-NOTFOUND")
-        message(FATAL_ERROR "
-python libraries not found, unable to build pycsdiff")
     endif()
```
kdudka, line 145: This `endif()` should be removed by this commit and re-introduced by the commit that needs it.  Otherwise the checkout of commit 9d9fa6f478ef649f59a284d228bccc366b1ef9b2 fails to build with:
```
CMake Error at src/CMakeLists.txt:145 (endif):
  Flow control statements are not properly nested.
```

kdudka (changes requested):
Thank you for working on it!  Some minor comments inline...

src/CMakeLists.txt
lzaoral: Thanks! Originally, I wanted to move the man page generation to a separate file (like in cswrap) but the target in the in `add_custom_target` has to be defined in the same file (otherwise you get errors) and this is a leftover from that.

lzaoral: Resolved in ce20660.

src/CMakeLists.txt
lzaoral: Resolved in 4539356.

lzaoral:
This rebased version contains a lot of new changes, mainly in the Python related part.

kdudka (approved):
Given the fact that we migrated production Covscan hub off RHEL-7 yesterday, the support for Python 2 is not much important any more.  Nevertheless, thank you for putting the effort into this.  The new build system logic can be reused for migration to Python 4 later on ;-)

lzaoral:
Thank you for the review, @kdudka!

PR #67, opened 2022-08-01
Fix `pycsdiff` build when more Python versions are used

lzaoral:


kdudka (approved):
Looks good but I would reword the commit messages to talk about Python **modules** rather than Python plugins.

kdudka (approved):

PR #68, opened 2022-08-01
Support ShellCheck SARIF rule metadata

### LINKED ISSUE #54: Support advanced ShellCheck rule descriptions for the SARIF format
With this, we could add e.g. an additional description of the given ShellCheck rule as we do in the HTML writer.

Additional information: https://github.com/microsoft/sarif-tutorials/blob/main/docs/2-Basics.md#rule-metadata

lzaoral:
This is a proof-of-concept at the moment.

Fixes #54.


lzaoral:
@jamacku, Can you please check that the version of `csgrep` from this branch creates correct help links for ShellCheck reports on GitHub? You can get the RPM from Packit. Thank you!

jamacku:
@lzaoral sure, I'll have a look. Thanks

jamacku (commented):
Hmm, from what I saw help paragraphs are still missing. I'm not sure where could be a problem.

image

lzaoral:
@jamacku, Could please try it again with the latest changes?  Note that this PR is just a proof-of-concept.  I'm not really happy with the current implementation.

jamacku:
@lzaoral It's unfortunate, but it still doesn't show a hint message in GitHub UI.

Screenshot from 2022-08-16 14-41-42

jamacku:
@lzaoral I came across this repo with SARIF examples: https://github.com/microsoft/sarif-tutorials

It seems like a good source of information if it's not outdated.

https://github.com/microsoft/sarif-tutorials/blob/main/docs/2-Basics.md#rule-metadata

lzaoral:
@jamacku, could you try it once again, please?

I've added the `fullDescription` key to the rule objects.  This key is mandatory accroding to the SARIF 2.1 standard. [1]   Note, that I've just copied a event message to this key for testing purposes, it will contain description that's not really general, e.g. `"COMMIT_TO_IM appears unused. Verify it or export it."`.

[1] https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317838

jamacku:
@lzaoral It WORKS :partying_face: 
Screenshot from 2022-08-22 16-29-50

But I'm not sure if it displays a correct message.

lzaoral:
@jamacku, Could you try it once more, please?  This time with much cleaner and simpler code.  Thanks in advance!

jamacku:
@lzaoral Please see:

- https://github.com/jamacku/differential-shellcheck/pull/29

Screenshot from 2022-08-24 09-05-55

It works great! :partying_face: I'm thinking if it would make sense to say something like:

```
Defect reference: https://github.com/koalaman/shellcheck/wiki/SC2086
```

Or maybe even try to take advantage of supported markdown templating and generate something like:

```markdown
Defect reference: [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086)
```

jamacku:
@lzaoral Works great :100: 

Markdown templating:

Screenshot from 2022-08-26 14-45-58

Defect searching by rule:

Screenshot from 2022-08-26 14-46-50

lzaoral:
This PR should be finally ready for a proper review. 🥳

jamacku (approved):
It works like a charm! LGTM :+1: 

- https://github.com/redhat-plumbers-in-action/differential-shellcheck/pull/117

kdudka (changes requested):
Thank you for moving it forward!  Could we please extend the SARIF writer such that CWE IDs are preserved when available even for ShellCheck?  We have our own mapping of `SCNNNN` to `CWE-NNN` in csmock: https://github.com/csutils/csmock/blob/0d96eeabbc064f828c35fbea95ebd3616295b8b7/cwe-map.csv#L425  I would like to avoid loosing this data when we convert csmock's output to SARIF.

lzaoral:
Unfortunately, SARIF allows at most one rule per defect.  Will be something like this enough?
```json
"properties": {
    "cwe": [
        "CWE-248"
    ]
    "tags": [
        "ShellCheck"
    ]
}
```

kdudka:
SARIF's rule is `def.checker + ": " + keyEvt.event` in the current output of `csgrep --mode=sarif`.  So assigning at most one rule to each defect is fine and we do not need to change it.  On the other hand, we can assign many properties to each rule.  I am not sure about the help text.  If there is just one help text field for each rule, we can probably concatenate the text and provide both the links?

lzaoral:
Yes, concatenating the help texts is possible.

lzaoral:
I've updated the PR with changes addressing concerns raised in https://github.com/csutils/csdiff/pull/68#pullrequestreview-1110911556.

kdudka (approved):
Thanks for the update!  I think it works as expected now. 

The implementation could be optimized for future extensions.  I would create a structure `RuleProps` and a single map from `ruleId` to `RuleProps`, in order to make adding new properties in the future more straightforward.  But this can be refactored later on when we actually need it.

jamacku (changes requested):
Hmm, it doesn't seem to work :crying_cat_face: 

Tested in:
- https://github.com/redhat-plumbers-in-action/differential-shellcheck/pull/204
[Reported defect](https://github.com/redhat-plumbers-in-action/differential-shellcheck/security/code-scanning/208) is missing metadata:

Screenshot from 2023-02-09 13-04-07

jamacku (approved):
When the defect was already detected before it won't have new metadata assigned to it, but when it is a fresh defect, it works as expected. It seems like a bug or implementation detail on the GitHub side.

Screenshot from 2023-02-09 13-14-07

LGTM :muscle:

PR #72, opened 2022-08-03
Port the project to C++11 - Part 2

lzaoral:
See commit messages.

### DIFF
--- a/src/instream.cc
+++ b/src/instream.cc
@@ -19,8 +19,8 @@
 
 #include "instream.hh"
 
-InStream::InStream(const std::string &fileName, const bool silent):
-    fileName_(fileName),
+InStream::InStream(std::string fileName, const bool silent):
+    fileName_(std::move(fileName)),
     silent_(silent),
     anyError_(false),
     str_((fileName_ == "-")
--- a/src/instream.cc
+++ b/src/instream.cc
@@ -41,12 +41,6 @@ InStream::InStream(std::istringstream &str, const bool silent):
 {
 }
 
-InStream::~InStream()
-{
-    if (&str_ == &fileStr_)
-        fileStr_.close();
-}
-
 void InStream::handleError(const std::string msg, const long line)
 {
     anyError_ = true;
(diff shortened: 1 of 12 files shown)


src/instream.cc
```diff
@@ -19,35 +19,27 @@
...
 }
 
-InStream::~InStream()
-{
-    if (&str_ == &fileStr_)
-        fileStr_.close();
-}
-
-void InStream::handleError(const std::string msg, const long line)
+void InStream::handleError(const std::string& msg, const unsigned long line)
```
kdudka, line 42: Please spell this as `const std::string &msg`.

src/instream.hh
```diff
@@ -30,37 +30,37 @@ struct InFileException {
     std::string fileName;
     // TODO: details (errno?)
 
-    InFileException(const std::string &fileName_):
-        fileName(fileName_)
+    InFileException(std::string fileName_):
+        fileName(std::move(fileName_))
```
kdudka, line 34: I fail to see how this would improve anything.  We save a copy of `std::string` in the constructor of `InFileException` at the cost of introducing a new copy of `std::string` at caller's side.  Let's not optimize something that has zero impact on performance.

kdudka (changes requested):
Thanks!  Overall it looks good.  Two minor comments inline...

src/instream.hh
lzaoral: I don't think I follow.  In the original code we always did one copy.  In the changed code, this is avoided in the case of temporaries (there will be moves but no unnecessary copies).

kdudka: The updated version does the copy here though (where the previous version only passed a reference): https://github.com/csutils/csdiff/blob/4b206d65e988f4225864b9894e9de06fa05a6eae/src/instream.cc#L34

src/instream.cc
lzaoral: Addressed in 80ea97f and another reason why #86 should be implemented as soon as possible. ;)

kdudka (approved):
Looks good.  Thanks!

lzaoral:
Thanks for the review, @kdudka!

Issue #80, opened 2022-08-26
Support `level` information in SARIF output


lzaoral:
We should also report the severity level [1].  Since we're not doing it at the moment, all reports default to `warning` and this property is also supported by GitHub [2].  Supported levels are: 

* `warning`
* `error`
* `note`
* `none`

[1] https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Toc10541086
[2] https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#result-object

PR #81, opened 2022-08-26
json-writer: print `level` key in SARIF output

lzaoral:
... so that users can easily filter results according to their severity.

Fixes: #80


src/json-writer.cc
```diff
@@ -214,6 +215,17 @@ static void sarifEncodeMsg(PTree *pDst, const std::string& text)
     pDst->put_child("message", msg);
 }
 
+static void sarifEncodeLevel(PTree *result, const std::string &event) {
+    std::string level = "warning";
+
+    if (boost::starts_with(event, "error"))
+        level = "error";
```
kdudka, line 222: I do not like the fact that each level is typed twice (in the match and in the assignment).  This is prone to typos as the list will grow.  Could we rather implement it along these lines?
```cxx
    std::string level = event;                                                                
                                                                                              
    // cut the [...] suffix from event if present                                             
    size_t pos = event.find('[');
    if (std::string::npos != pos)
        level = event.substr(0U, pos);

    // go through events that denote warning level                                            
    for (const char *str : {"error", "warning", "note"}) {                                    
        if (str == level) {                                                                   
            // encode in the output if matched                                                
            result->put<std::string>("level", level);
            return;                                                                           
        }
    }   
```

kdudka (changes requested):
Thank you for working on it!  Some minor issues inline...

src/json-writer.cc
lzaoral: That looks perfect, thanks!

lzaoral: Addressed in a3eac76.

src/json-writer.cc
```diff
@@ -214,6 +215,17 @@ static void sarifEncodeMsg(PTree *pDst, const std::string& text)
     pDst->put_child("message", msg);
 }
 
+static void sarifEncodeLevel(PTree *result, const std::string &event) {
+    std::string level = "warning";
```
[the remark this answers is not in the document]
lzaoral: Addressed in a3eac76.

kdudka (approved):
Thanks for the update!

jamacku (approved):
This is amazing, Thank you! :partying_face: 

Screenshot from 2022-08-26 21-19-53

src/json-writer.cc
```diff
@@ -214,6 +214,24 @@ static void sarifEncodeMsg(PTree *pDst, const std::string& text)
     pDst->put_child("message", msg);
 }
 
+static void sarifEncodeLevel(PTree *result, const std::string &event) {
```
lzaoral, line 217: Note to self: fix bracket formatting inconsistency in some future commit

kdudka: We might consider using this one-liner in CI to catch this:
```
grep '^static .*) *{$' src/*
```

There is one more inconsistency in `abstarct-parser.cc`:
```
src/abstract-parser.cc:static inline std::unique_ptr<T> make_unique(InStream &input) {
src/json-writer.cc:static void sarifEncodeLevel(PTree *result, const std::string &event) {
```

lzaoral: Using a custom `.clang-format` config to enforce code style is would be a better solution.
tracked in https://github.com/csutils/csdiff/issues/86.

kdudka: That would mean more work than the proposed one-liner.  But if we have a volunteer to work on this, why not :-)

lzaoral: It's not that hard, I've already done it few times.  Additionally, such config file can be then used in all of our projects.

Issue #75, opened 2022-08-22
Please add support for ShellCheck JSON format


jamacku:
ShellCheck can suggest a possible solution of the reported defect. Unfortunately, this data isn't available in `gcc` format. It would be nice if `csdiff` could parse ShellCheck JSON format, which contains possible fix under `fix` key.

[Example](https://www.shellcheck.net/wiki/Integration#GCC-compatible-error-messages):

```json
{
  "comments": [
    {
      "file": "myotherscript",
      "line": 2,
      "endLine": 2,
      "column": 1,
      "endColumn": 2,
      "level": "error",
      "code": 1035,
      "message": "You need a space after the [ and before the ].",
      "fix": null
    },
    {
      "file": "myscript",
      "line": 2,
      "endLine": 2,
      "column": 6,
      "endColumn": 8,
      "level": "warning",
      "code": 2039,
      "message": "In POSIX sh, echo flags are undefined.",
      "fix": null
    },
    {
      "file": "myscript",
      "line": 2,
      "endLine": 2,
      "column": 9,
      "endColumn": 11,
      "level": "info",
      "code": 2086,
      "message": "Double quote to prevent globbing and word splitting.",
      "fix": {
        "replacements": [
          {
            "line": 2,
            "endLine": 2,
            "precedence": 7,
            "insertionPoint": "afterEnd",
            "column": 9,
            "replacement": "\"",
            "endColumn": 9
          },
          {
            "line": 2,
            "endLine": 2,
            "precedence": 7,
            "insertionPoint": "beforeStart",
            "column": 11,
            "replacement": "\"",
            "endColumn": 11
          }
        ]
      }
    }
  ]
}
```

Having this kind of information would be very useful for:
- https://github.com/redhat-plumbers-in-action/differential-shellcheck/issues/88

Thanks for working on this project! :+1:

kdudka:
I tried to run `shellcheck --format=json **/*.sh` and the output looked like this (the top-level `"comments":` was not there):
```json
[
  {
    "file": "tests/test-lib.sh",
    "line": 1,
    "endLine": 1,
    "column": 1,
    "endColumn": 1,
    "level": "error",
    "code": 2148,
    "message": "Tips depend on target shell and yours is unknown. Add a shebang or a 'shell' directive.",
    "fix": null
  },
  {
    "file": "tests/cssort/sync-diff.sh",
    "line": 2,
    "endLine": 2,
    "column": 8,
    "endColumn": 14,
[...]

PR #83, opened 2022-09-08
json-parser: basic support for ShellCheck's JSON output

kdudka:
... which is produced by `shellcheck --format=json ...`


jamacku (changes requested):
@kdudka Thank you. Only one small thing. I think the commit message should say `--format=json1` instead of only `json` as mentioned in:

- https://github.com/csutils/csdiff/issues/75

kdudka:
@jamacku Good catch!  Fixed.

lzaoral (changes requested):
The code looks great but I'd drop the `Fixes:` from the commit message.  The issue was specifically asking about handling of ShellCheck's suggestions which are still marked as a TODO.

kdudka:
@lzaoral You are right.  Fixed.

lzaoral (approved):
Thanks, @kdudka!

jamacku (approved):

kdudka:
@lzaoral Note that using `Related: ` in the commit message does not prevent GitHub from closing the issue.

PR #89, opened 2022-11-01
cstrans-df-run: simplify the code using C++11 extensions

kdudka:
We can use `std::ifstream` and `std::ofstream` to simplify construction of the stream objects.  And we do not need to close the streams when they are going to be destroyed (and hence automatically closed).


lzaoral (approved):
LGTM! Thank you, @kdudka!

Issue #90, opened 2022-11-02
json-parser: support the JSON format produced by RapiDAST


kdudka:
The JSON parser used by csdiff utilities should accept the JSON format produced by [RapiDAST](https://github.com/RedHatProductSecurity/rapidast/), which internally uses [OWASP ZAP Proxy](https://owasp.org/www-project-zap/).  Two examples of the results format follow.

`vapi-report-20220809-143815.json`:
```json
{
        "@version": "2.11.1",
        "@generated": "Tue, 9 Aug 2022 14:38:31",
        "site":[ 
                {
                        "@name": "http://rhos-fedora-devel.usersys.redhat.com:5000",
                        "@host": "rhos-fedora-devel.usersys.redhat.com",
                        "@port": "5000",
                        "@ssl": "false",
                        "alerts": [ 
                                {
                                        "pluginid": "90022",
                                        "alertRef": "90022",
                                        "alert": "Application Error Disclosure",
                                        "name": "Application Error Disclosure",
                                        "riskcode": "1",
                                        "confidence": "2",
                                        "riskdesc": "Low (Medium)",
                                        "desc": "<p>This page contains an error/warning message that may disclose sensitive information like the location of the file that produced the unhandled exception. This information can be used to launch further attacks against the web application. The alert could be a false positive if the error message is found inside a documentation page.</p>",
                                        "instances":[ 
                                                {
                                                        "uri": "http://rhos-fedora-devel.usersys.redhat.com:5000/pets/id/pet_id",
                                                        "method": "GET",
                                                        "param": "",
                                                        "attack": "",
                                                        "evidence": "HTTP/1.1 500 INTERNAL SERVER ERROR"
                                                }
                                        ],
                                        "count": "1",
                                        "solution": "<p>Review the source code of this page. Implement custom error pages. Consider implementing a mechanism to provide a unique error reference/identifier to the client (browser) while logging the details on the server side and not exposing them to the user.</p>",
                                        "otherinfo": "",
                                        "reference": "",
                                        "cweid": "200",
                                        "wascid": "13",
                                        "sourceid": "7"
                                },
                                {
                                        "pluginid": "10023",
                                        "alertRef": "10023",
                                        "alert": "Information Disclosure - Debug Error Messages",
                                        "name": "Information Disclosure - Debug Error Messages",
                                        "riskcode": "1",
                                        "confidence": "2",
                                        "riskdesc": "Low (Medium)",
[...]

PR #92, opened 2022-11-04
abstract-tree: let decoder iterate over tree nodes on its own

kdudka:
Before this change, iteration over tree nodes was handled externally but it only worked when all the visited nodes were at the same level.

In order to handle the JSON output format produced by OWASP ZAP, we need to iterate over two levels:
1. The top-level "site" array.
2. The nested "alerts" array.

To handle tree formats like this, it works better when the decoder iterates over the nodes on its own, each time the `readNode()` method is called.

No change in behavior is intended by this commit.

Related: https://github.com/csutils/csdiff/issues/90

### DIFF
--- a/src/json-parser.cc
+++ b/src/json-parser.cc
@@ -40,7 +40,7 @@ class SimpleTreeDecoder: public AbstractTreeDecoder {
                 const pt::ptree        *root)
             override;
 
-        bool readNode(Defect *def, pt::ptree::const_iterator defIter) override;
+        bool readNode(Defect *def) override;
 
     private:
         enum ENodeKind {
@@ -63,7 +63,7 @@ class SimpleTreeDecoder: public AbstractTreeDecoder {
 /// tree decoder of the Coverity JSON format
 class CovTreeDecoder: public AbstractTreeDecoder {
     public:
-        bool readNode(Defect *def, pt::ptree::const_iterator defIter) override;
+        bool readNode(Defect *def) override;
 
     private:
         KeyEventDigger              keDigger;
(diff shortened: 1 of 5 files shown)


src/json-parser.cc
```diff
@@ -314,11 +287,15 @@ void SimpleTreeDecoder::readScanProps(
...
+bool SimpleTreeDecoder::readNode(Defect *def)
 {
-    const pt::ptree &defNode = defIter->second;
+    if (!defList_ || defList_->end() == defIter_)
+        // failed initialization or EOF
+        return false;
+
+    // move the iterator after we get the current position
+    const pt::ptree::const_iterator itNow = defIter_++;
+    const pt::ptree &defNode = itNow->second;
```
lzaoral, line 298: I know that this code is not general for every Decoder instance, but it's duplicated few times so I suggest we should factor it out.

lzaoral (changes requested):
Apart from the comment inline, LGTM and thanks for the C++11 stuff!

src/json-parser.cc
kdudka: Thanks for review!  Fixed in commit 8460c17383db4c2e8db64a0a90fb23642df5cc0a.

lzaoral (approved):

lzaoral (changes requested):
Sorry for a second round, could you also remove these constructors:
```
 JsonParser(const Parser &);
        JsonParser& operator=(const Parser &);
```
They are redundant since `std::unique_ptr` cannot be copied. Same goes for the `XmlParser`. Thank you!

kdudka:
Good point!  Fixed...

lzaoral (approved):
LGTM! Thanks, @kdudka!

PR #100, opened 2023-01-17
src: Add ShellCheck directive to fix: error[SC2148]

jamacku:
Fixes:
- https://github.com/csutils/csdiff/security/code-scanning/48


kdudka (approved):
Merging, thanks!

PR #103, opened 2023-01-30
Port the project to C++11 - Part 3

lzaoral:
See commit messages.

### DIFF
--- a/src/lib/csv-parser.cc
+++ b/src/lib/csv-parser.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020 Red Hat, Inc.
+ * Copyright (C) 2020-2023 Red Hat, Inc.
  *
  * This file is part of csdiff.
  *
@@ -49,8 +49,8 @@ bool AbstractCsvParser::parse(InStream &ins)
     std::string line;
     for (d->lineno = 1; std::getline(str, line); d->lineno++) {
         // initialize tokenizer
-        typedef boost::escaped_list_separator<char>     TSeparator;
-        typedef boost::tokenizer<TSeparator>            TTokenizer;
+        using TSeparator = boost::escaped_list_separator<char>;
+        using TTokenizer = boost::tokenizer<TSeparator>;
         TTokenizer tok(line);
 
         // break the current line into fields
--- a/src/lib/csv-parser.cc
+++ b/src/lib/csv-parser.cc
@@ -34,9 +34,7 @@ AbstractCsvParser::AbstractCsvParser():
 {
 }
 
-AbstractCsvParser::~AbstractCsvParser()
-{
-}
+AbstractCsvParser::~AbstractCsvParser() = default;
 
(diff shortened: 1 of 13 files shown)


src/lib/msg-filter.hh
```diff
@@ -32,12 +33,13 @@ using TSubstMap = std::map<std::string, std::string>;
 class MsgFilter {
     public:
         // singleton
-        static MsgFilter* inst() {
-            return (self_)
-                ? (self_)
-                : (self_ = new MsgFilter);
+        static MsgFilter &inst() {
```
kdudka, line 36: I would prefer to spell this as `static MsgFilter& inst()` as we do in other function declarations.

src/lib/msg-filter.hh
```diff
@@ -47,18 +49,19 @@ class MsgFilter {
...
-        ~MsgFilter();
 
-        bool setJSONFilter(InStream &filter);
+        bool setJSONFilter(InStream &input);
 
-        static MsgFilter *self_;
         struct Private;
-        Private *d;
+        std::unique_ptr<Private> d;
+        static std::unique_ptr<MsgFilter> self_;
```
kdudka, line 64: I am not sure whether we want to use `std::unique_ptr` for the global variable.  It introduces one more thing to debug on program exit.  Is the motivation behind the change to have clean valgrind output?

src/lib/csv-parser.cc
```diff
@@ -34,10 +34,7 @@ AbstractCsvParser::AbstractCsvParser():
 {
 }
 
-AbstractCsvParser::~AbstractCsvParser()
-{
-    delete d;
-}
+AbstractCsvParser::~AbstractCsvParser() = default;
```
kdudka, line 37: I propose to squash commits 0b773b27f3e916cba754c146d2c57ff92ed76806 and fc483a65105ca1d636cde9965608f3e452ee3e95.  As it is now, the second commit fixes a memory leak that the first one introduces.

kdudka (approved):
Looks good. Minor comments inline...

src/lib/msg-filter.hh
lzaoral: The main motivation was consistency. But I have no strong feelings about this one. Should I remove this change?

src/lib/msg-filter.hh
lzaoral: Addressed in 78b43f890cbd4fcd2b7843a519f0fe2f9cffa758.

src/lib/csv-parser.cc
lzaoral: Thanks! I've mixed those two by a mistake. Addressed in a92210518f357a3a2a4f7aa6b634452a5a556c3e and e419c57a39b15090d71d3be6619d197665ae996e.

src/lib/msg-filter.hh
kdudka: Yes, please.  I would prefer to use a raw pointer for the static variable.

PR #111, opened 2023-02-27
Set UBSAN options using CMake

lzaoral:


lzaoral:
TIL, that `set_tests_properties` does not allow appending so that was the reason why the Ci was failing.

kdudka (approved):

PR #105, opened 2023-01-30
Rewrite `writer-json` using Boost.JSON

lzaoral:
CMake changes expect a build of Boost with Boost.JSON.

PR #115, opened 2023-03-04
Json writer: use separate compilation units for each JSON scheme

kdudka:


src/lib/writer-json-common.cc
```diff
@@ -0,0 +1,150 @@
...
+    return convert_string<char>(str.data(), str.data() + str.size());
+}
+
+// TODO: This should not necessary!  TScanProps should be able to contain
+// any type so that no conversions here are needed.
+boost::json::object jsonSerializeScanProps(const TScanProps &scanProps)
+{
+    static auto isDigit = [](unsigned char c){ return std::isdigit(c); };
+
+    boost::json::object scan;
```
lzaoral, line 42: Previous line defines `using namespace boost::json;` so we can use:
```suggestion
object jsonSerializeScanProps(const TScanProps &scanProps)
{
    static auto isDigit = [](unsigned char c){ return std::isdigit(c); };

    object scan;
```

src/lib/writer-json-sarif.hh
```diff
@@ -0,0 +1,61 @@
...
+        using TCweMap = std::map<std::string, int>;
+        TCweMap                     cweMap_;
+
+        using TShellCheckMap = std::map<std::string, std::string>;
+        TShellCheckMap              shellCheckMap_;
+
+        TScanProps                  scanProps_;
+        object                      driver_;
+        array                       results_;
+};
```
lzaoral, line 59: I don't think it's a good idea to flatten a namespace in a header file.  The using declaration should be moved to `src/lib/writer-json-sarif.cc` instead.

lzaoral (changes requested):
LGTM apart from the `using namespace` in a header!  Details inline.

src/lib/writer-json-common.cc
kdudka: Fixed in commit 184c906270a36e8c823cdc4d1f551d8dfd676170.

src/lib/writer-json-sarif.hh
kdudka: Right.  It was a copy-paste mistake.  Thank you for spotting it!

kdudka: Fixed in commit 71c1aee21c380e3b647ebfba5d94802a80de2247.

src/lib/writer-json-common.cc
```diff
@@ -0,0 +1,150 @@
...
+        for (const auto &elem : arr) {
+            os << sep << *indent;
+            jsonPrettyPrint(os, elem, indent);
+            sep = ",\n";
+        }
+        os << '\n';
+
+        indent->resize(indent->size() - 4);
+        os << *indent << ']';
+
```
lzaoral, line 77: Redundant empty line. Sorry for introducing it in #105.

src/lib/writer-json-common.cc
```diff
@@ -0,0 +1,150 @@
...
+        os << '\n';
+
+        indent->resize(indent->size() - 4);
+        os << *indent << '}';
+}
+
+void jsonPrettyPrint(
+        std::ostream               &os,
+        const boost::json::value   &jv,
+        std::string                *indent)
```
lzaoral, line 108:
```suggestion
        std::ostream               &os,
        const value                &jv,
        std::string                *indent)
```

lzaoral (approved):
LGTM, thanks @kdudka!  Minor nits I missed yesterday are inline.

src/lib/writer-json-common.cc
kdudka: Fixed in commit c436c719f89e7f071c423c5e658a42b6b48a6801.

src/lib/writer-json-common.cc
kdudka: Fixed in commit c436c719f89e7f071c423c5e658a42b6b48a6801.

kdudka:
@lzaoral Thanks for review!  Merging...

PR #117, opened 2023-03-13
Replace Boost.Regex with the standard library

lzaoral:
We removed the hard dependency on Boost.Regex due to the usage of regex streams when we switched to Boost.JSON in #105.  Let's use what the standard library provides instead.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -38,7 +38,6 @@ endif()
 find_package(Boost REQUIRED
     COMPONENTS filesystem
                program_options
-               regex
                system
     OPTIONAL_COMPONENTS json
                         nowide)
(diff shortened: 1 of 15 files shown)


lzaoral:
Changing to a draft since EPEL 7 does not provide all `std::regex_replace` variants [1].  Especially those where `fmt` is a string literal.

[1] https://en.cppreference.com/w/cpp/regex/regex_replace

lzaoral:
Yeah, this definitely won't work as the regex library was implemented in GCC 4.9 [1] which is newer than GCC in EPEL 7. 

[1] https://en.cppreference.com/w/cpp/compiler_support/11#C.2B.2B11_library_features

lzaoral:
@kdudka Please, let me know in the future when `csdiff` drops support for RHEL 7 so that I can rebase and finish this PR.  Thank you!

lzaoral:
Actually, `csdiff` now uses named capturing groups in some regular expression and this feature is not supported by any C++ standard.  Therefore, this PR is obsolete.

PR #121, opened 2023-04-05
csgrep --warning-rate-limit: fix msg of the note event

kdudka:
The code used a string that was already overwritten for the error event.


lzaoral (approved):
Nice catch!

kdudka:
@lzaoral Thanks for review!

PR #123, opened 2023-04-12
writer-json-sarif: unconditionally set the `region` property

kdudka:
... for each `physicalLocation` object.  Although the property is optional according to the specification of SARIF Version 2.1.0, DefectDojo fails to import the data if the property is missing.

Reported-by: Mert Bugra Bicak
Closes: https://github.com/csutils/csdiff/pull/123

### DIFF
--- a/src/lib/writer-json-sarif.cc
+++ b/src/lib/writer-json-sarif.cc
@@ -212,16 +212,14 @@ static void sarifEncodeLoc(object *pLoc, const Defect &def, unsigned idx)
     };
 
     // line/col
-    if (evt.line) {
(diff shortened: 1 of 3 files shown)


lzaoral (approved):
LGTM!  Thanks, @kdudka!

kdudka:
@lzaoral Thanks for review!

kdudka:
I discovered today that this change broke the validation because `startLine` cannot be zero:
```
runs[0].results[0].relatedLocations[0].physicalLocation.region.startLine: 0 is less than the minimum permitted value of 1.
```

Issue #125, opened 2023-04-17
encode CWEs in per-results properties, too, in SARIF output


kdudka:
As originally reported by Mert Bugra Bicak, the SARIF writer in csdiff should encode CWEs in per-results properties, too, so that the SARIF parser in DefectDojo can process them: `result.properties.cwe[ ]`

mertbugrabicak:
Just a quick adjustment to this one after testing it out on DefectDojo: It seems to require only one CWE value per finding and gives an error when trying to do result.properties.cwe[] with an array of strings.

Instead in must be results.properties.cwe with just a single CWE value as a string.

This seems to be due to the limitation on only being able to have 1 CWE per finding in DefectDojo. So if a finding in our results has multiple CWEs, we would need to narrow it down to 1 for it to be successfully imported

PR #126, opened 2023-04-17
writer-json-sarif: encode per-warning CWE property

kdudka:
... so that the data is processed by DefectDojo

Fixes: https://github.com/csutils/csdiff/issues/125
Closes: https://github.com/csutils/csdiff/pull/126

### DIFF
--- a/src/lib/writer-json-sarif.cc
+++ b/src/lib/writer-json-sarif.cc
@@ -276,9 +276,14 @@ void SarifTreeEncoder::appendDef(const Defect &def)
         shellCheckMap_[ruleId] = sm[2];
     }
 
-    if (def.cwe)
+    if (def.cwe) {
         // update CWE map
         cweMap_[ruleId] = def.cwe;
+
+        // encode per-warning CWE property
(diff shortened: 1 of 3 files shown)


src/lib/writer-json-sarif.cc
```diff
@@ -276,10 +276,15 @@ void SarifTreeEncoder::appendDef(const Defect &def)
         shellCheckMap_[ruleId] = sm[2];
     }
 
-    if (def.cwe)
+    if (def.cwe) {
         // update CWE map
         cweMap_[ruleId] = def.cwe;
 
+        // encode per-warning CWE property
+        const object cweProp = {{ "cwe", "CWE-" + std::to_string(def.cwe) }};
```
lzaoral, line 284: You can't move a `const` object because the type of `std::move(cweProp)` is `const object&&` which is not a proper candidate for the move assignment operator.
```suggestion
        object cweProp = {{ "cwe", "CWE-" + std::to_string(def.cwe) }};
```

lzaoral (approved):
Minor performance issue inline, otherwise LGTM!

Issue #124, opened 2023-04-17
provide code snippet for key event location explicitly in SARIF output


kdudka:
As originally reported by Mert Bugra Bicak, the SARIF writer in csdiff should provide code snippet for key event location explicitly, as in this example: https://github.com/DefectDojo/django-DefectDojo/blob/master/unittests/scans/sarif/gitleaks_7.5.0.sarif  so that the SARIF parser in DefectDojo can process them.

kdudka:
@mertbugrabicak Could you please try csdiff packages from the following COPR and check whether they produce the expected output?  https://copr.fedorainfracloud.org/coprs/packit/csutils-csdiff-127/

mertbugrabicak:
Hi @kdudka , just checked and it seems to be working, thanks! Both the CWEs and the code snippet are included in the imported results now

I only have a tiny thing to point out: Before concatenating the snippet lines like LINE1\nLINE2\n... can we strip each line of their leading spaces? I think when DefectDojo imports that snippet text, it seems to lose the leading spaces in the first line, and it causes a right-shift in the continuing lines, causing it to look a little mis-aligned:

For example, in DefectDojo it looks like
1269|
  1270|
  1271|->
  1272|
  1273|

Where if we strip each line before concatenating, it can look align as below:
1269|
1270|
1271|->
1272|
1273|

This is purely aesthetics of course

mertbugrabicak:
Oops github also seems to strip the leading spaces from my example :)

kdudka:
@mertbugrabicak Thank you for testing it!

> Before concatenating the snippet lines like LINE1\nLINE2\n... can we strip each line of their leading spaces?

The problem is that it would still be misaligned if the code snippet was spread across line numbers with different count of digits:
```
98|
99|
100|->
101|
102|
```

I am wondering whether we could prepend some special character to prevent DefectDojo from stripping the leading white spaces?

mertbugrabicak:
Oh I hadn't thought of those types of cases, that's true. Yeah, I think that might work

kdudka:
I attempted to fix this in commit de3c5c97406e4889c07ceb53c4fc714bd84cb351.  @mertbugrabicak Could you please update to `csdiff-3.0.1.20230420.085832.gde3c5c9.pr_127-1` and check whether it works any better?

mertbugrabicak:
@kdudka Just checked and it looks great, fixes the problem, thank you!

kdudka:
[...]

PR #127, opened 2023-04-19
writer-json-sarif: encode code snippets for DefectDojo

kdudka:
... in the same way as Snyk Code encodes them

Fixes: https://github.com/csutils/csdiff/issues/124
Closes: https://github.com/csutils/csdiff/pull/127

### DIFF
--- a/src/lib/writer-json-sarif.cc
+++ b/src/lib/writer-json-sarif.cc
@@ -258,6 +258,27 @@ static void sarifEncodeEvt(array *pDst, const Defect &def, unsigned idx)
     pDst->push_back(std::move(tfLoc));
 }
 
+void sarifEncodeSnippet(object &reg, const std::string &msg)
+{
+    // check whether the "snippet" node already exists
+    value &valSnip = reg["snippet"];
+    if (!valSnip.is_object()) {
+        // create the "text" node containing the header line
+        valSnip.emplace_object() = {
+            { "text", "Problem detected in this context:" }
+        };
+    }
+
+    // reuse the existing "snippet/text" nodes
+    string &strSnip = valSnip.get_object()["text"].get_string();
+
+    // use new-line as delimiter
+    strSnip += "\n";
+
+    // concatenate the snippet from this event
+    strSnip += msg;
+}
+
 void SarifTreeEncoder::appendDef(const Defect &def)
 {
     const DefEvent &keyEvt = def.events[def.keyEventIdx];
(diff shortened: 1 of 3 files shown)


src/lib/writer-json-sarif.cc
```diff
@@ -258,6 +258,26 @@ static void sarifEncodeEvt(array *pDst, const Defect &def, unsigned idx)
...
 }
 
+void sarifEncodeSnippet(object *reg, const std::string &msg)
+{
+    // check whether the "snippet" node already exists
+    value &valSnip = (*reg)["snippet"];
+    if (!valSnip.is_object()) {
+        // create the "text" node containing the header line
+        object objText = {{ "text", "Problem detected in this context:" }};
+        valSnip = objText;
```
lzaoral, line 268: You can use `emplaceObject` to create and access a object in place: https://www.boost.org/doc/libs/1_79_0/libs/json/doc/html/json/ref/boost__json__value/emplace_object.html

```suggestion
        valSnip.emplaceObject() = {{ "text", "Problem detected in this context:" }};
```

src/lib/writer-json-sarif.cc
```diff
@@ -293,16 +313,30 @@ void SarifTreeEncoder::appendDef(const Defect &def)
     sarifEncodeLoc(&loc, def, def.keyEventIdx);
     result["locations"] = array{std::move(loc)};
 
+    // resolve key event region
+    value &valLoc = result["locations"].get_array().front();
+    value &valPhy = valLoc.get_object()["physicalLocation"];
+    object *reg = &valPhy.get_object()["region"].get_object();
```
lzaoral, line 319: I'd create a new `reg` object here and then I'd move it to `loc[physicalLocation]` and then move `loc` to `result["locations"]`.

lzaoral (changes requested):

src/lib/writer-json-sarif.cc
kdudka: Addressed in commit 3adeff9a317a808cb932c51bbc809676db06b62e.

src/lib/writer-json-sarif.cc
kdudka: As discussed in person, this would reorder object properties in the output, which would imply additional changes of the expected output in the test-suite.

kdudka:
@lzaoral Thanks for review!

Issue #108, opened 2023-02-01
Support UBSAN reports


lzaoral:
Example output:
```
byteorder.h:83:9: runtime error: load of misaligned address 0x556e3e877805 for type 'const uint32_t', which requires 4 byte alignment
0x556e3e877805: note: pointer points here
 b5 21 00 00 6c 00 00  07 ff 65 a0 b8 03 05 2f  74 65 78 74 0e 70 d6 f0  d2 4d 97 21 a4 81 00 00  a0
             ^
    #0 0x556e3dc9349f in IVALu /builddir/build/BUILD/rsync-3.2.3/byteorder.h:83
    #1 0x556e3dc9349f in IVAL /builddir/build/BUILD/rsync-3.2.3/byteorder.h:124
    #2 0x556e3dc9349f in raw_read_int /builddir/build/BUILD/rsync-3.2.3/io.c:921
    #3 0x556e3dc9349f in read_a_msg /builddir/build/BUILD/rsync-3.2.3/io.c:1441
    #4 0x556e3dc93b23 in read_buf /builddir/build/BUILD/rsync-3.2.3/io.c:1853
    #5 0x556e3dc958d1 in read_ndx /builddir/build/BUILD/rsync-3.2.3/io.c:2241
    #6 0x556e3dc31316 in read_ndx_and_attrs /builddir/build/BUILD/rsync-3.2.3/rsync.c:330
    #7 0x556e3dc43f51 in recv_files /builddir/build/BUILD/rsync-3.2.3/receiver.c:548
    #8 0x556e3dc664b2 in do_recv /builddir/build/BUILD/rsync-3.2.3/main.c:1048
    #9 0x556e3dc66fc7 in do_server_recv /builddir/build/BUILD/rsync-3.2.3/main.c:1219
    #10 0x556e3dc66fc7 in start_server /builddir/build/BUILD/rsync-3.2.3/main.c:1253
    #11 0x556e3dc67418 in child_main /builddir/build/BUILD/rsync-3.2.3/main.c:1226
    #12 0x556e3dcca5f2 in local_child /builddir/build/BUILD/rsync-3.2.3/pipe.c:166
    #13 0x556e3dc0bb33 in do_cmd /builddir/build/BUILD/rsync-3.2.3/main.c:650
    #14 0x556e3dc0bb33 in start_client /builddir/build/BUILD/rsync-3.2.3/main.c:1576
[...]

PR #129, opened 2023-05-29
parser-gcc: implement parsing of UBSAN reports

lzaoral:
Resolves: https://github.com/csutils/csdiff/issues/108


src/lib/parser-gcc.cc
```diff
@@ -155,17 +166,22 @@ EToken Tokenizer::readNext(DefEvent *pEvt)
         pEvt->msg   = sm[/* fnc */ 4] + "(): ";
         pEvt->msg  += sm[/* msg */ 6];
     }
+    else if (boost::regex_match(line, sm, reUbsanScope_)) {
+        tok = T_MSG;
+        pEvt->event = "note";
+        pEvt->msg   = sm[/* fnc */ 2] + " at " + sm[/* address */ 1];
```
kdudka, line 172: Should we add the `()` suffix after `fnc`, as we do for smatch?

src/lib/parser-gcc.cc
```diff
@@ -741,7 +762,8 @@ bool GccParser::Private::tryMerge(Defect *pDef)
         // not something we can merge
         return false;
 
-    if (pDef->checker != this->lastDef.checker && lastKeyEvt.event != "#")
+    if (pDef->checker != this->lastDef.checker && lastKeyEvt.event != "#"
+            && pDef->checker != "UBSAN_WARNING")
```
kdudka, line 766: It is not obvious why we need to special-case `UBSAN_WARNING` here.  Could we please cover it by a test-case?

src/lib/parser-gcc.cc
```diff
@@ -496,6 +512,8 @@ bool BasicGccParser::exportAndReset(Defect *pDef)
         def.checker = "SHELLCHECK_WARNING";
     else if (boost::regex_match(keyEvt.msg, reSmatchMsg_))
         def.checker = "SMATCH_WARNING";
+    else if (keyEvt.event == "runtime error")
```
kdudka, line 515: Is `runtime error` sufficiently specific to `UBSAN_WARNING`?  Do we have a plan for future extensions?  We should probably clarify it in a comment.

kdudka (changes requested):
Thank you for working on this!  Although the line diff looks small, the single commit combines multiple more or less independent changes.  I would split it into 3 smaller commits:
- change the way how we refer to sub-expressions in regular expressions (no change in behavior)
- make the GCC parser silently ignore empty lines (change in behavior not specific to UBSAN)
- finally the changes specific to the UBSAN data format

src/lib/parser-gcc.cc
lzaoral: I'll add a comment explaining it.  The line is already covered by a test, without this change, the `note` events would not be merged with the `runtime error` key event.

src/lib/parser-gcc.cc
lzaoral: Good idea!  Addressed in 484354606f7b4fe877235831f7cd1c5e7e82a5da.

src/lib/parser-gcc.cc
lzaoral: Comment added in 484354606f7b4fe877235831f7cd1c5e7e82a5da.

src/lib/parser-gcc.cc
lzaoral: Comment added in 484354606f7b4fe877235831f7cd1c5e7e82a5da.

kdudka: This was indeed covered by a test but the test was not included in the cmake list when I was doing the review, so the test-suite was passing by mistake for me.  I think a saner solution would be to move the matching of `UBSAN_WARNING` to `GccPostProcessor` so that we do not need to special-case `UBSAN_WARNING` in the defect merging loop.  I have proposed the solution in commit 8a5fcb94fa5e76141f1c25bb7c46f371537b925a.

kdudka (approved):
Looks good.  Thanks for the contribution!

PR #130, opened 2023-05-31
parser-gcc: fix incomplete file paths in UBSAN key events

lzaoral:
UBSAN may only use the base name for the file path in its key event. However, the top of the backtrace, if present, always contains the whole path to this file.

This fixes `csmock` throwing all UBSAN warnings into the suppressed category.

### DIFF
--- a/src/lib/parser-gcc.cc
+++ b/src/lib/parser-gcc.cc
@@ -23,6 +23,7 @@
 #include "regex.hh"
 
 #include <algorithm>
+#include <boost/filesystem.hpp>
 
 namespace GccParserImpl {
 
@@ -639,8 +640,32 @@ void GccPostProcessor::Private::transUbsan(Defect *pDef) const
 
     // UBSAN always uses 'runtime error' for its key event
     DefEvent &keyEvt = pDef->events[pDef->keyEventIdx];
-    if (keyEvt.event == "runtime error")
-        pDef->checker = "UBSAN_WARNING";
+    if (keyEvt.event != "runtime error")
+        return;
+
+    pDef->checker = "UBSAN_WARNING";
+
+    // UBSAN may only use the base name for the file path in its key event.
+    // However, the top of the backtrace, if present, always contains the
+    // whole path to this file.
+    if (!keyEvt.fileName.empty() && keyEvt.fileName.front() == '/')
+        return;
+
+    using path = boost::filesystem::path;
+    const auto &keyFileName = path(keyEvt.fileName).filename();
+
+    for (const auto &evt : pDef->events) {
+        const auto &evtFileName = path(evt.fileName).filename();
+
+        // check that the basenames and line numbers match
+        if (&keyEvt != &evt
+                && keyFileName == evtFileName
+                && keyEvt.line == evt.line)
+        {
+            keyEvt.fileName = evt.fileName;
+            break;
+        }
+    }
 }
 
 void GccPostProcessor::Private::polishGccAnal(Defect *pDef) const
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -57,6 +57,7 @@ include_directories(lib)
 
 # link cslib.a and boost libraries
 link_libraries(cs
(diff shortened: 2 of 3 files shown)


src/lib/parser-gcc.cc
```diff
@@ -641,6 +641,17 @@ void GccPostProcessor::Private::transUbsan(Defect *pDef) const
     DefEvent &keyEvt = pDef->events[pDef->keyEventIdx];
     if (keyEvt.event == "runtime error")
         pDef->checker = "UBSAN_WARNING";
+
```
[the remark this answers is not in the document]
lzaoral: Addressed in c2c1ae32447d01edb93c5e9c48bddc186785b692.

src/lib/parser-gcc.cc
```diff
@@ -639,8 +639,21 @@ void GccPostProcessor::Private::transUbsan(Defect *pDef) const
...
+    pDef->checker = "UBSAN_WARNING";
+
+    // UBSAN may only use the base name for the file path in its key event.
+    // However, the top of the backtrace, if present, always contains the
+    // whole path to this file.
+    for (const auto &evt : pDef->events) {
+        size_t btAt = evt.msg.find("() at ");
+        if (std::string::npos != btAt) {
+            keyEvt.fileName = evt.fileName;
```
kdudka, line 653: Should we check that the line numbers (and maybe basenames?) match for the pair of events so that we do not combine file name and position from two different locations (e.g. when function inlining is involved)?

lzaoral: Addressed in 4bf72145084eaaae77e5d8301bb615b4e0b26e27.  Sorry for such a late response!

src/lib/parser-gcc.cc
```diff
@@ -639,8 +640,26 @@ void GccPostProcessor::Private::transUbsan(Defect *pDef) const
...
+    // UBSAN may only use the base name for the file path in its key event.
+    // However, the top of the backtrace, if present, always contains the
+    // whole path to this file.
+    if (!keyEvt.fileName.empty() && keyEvt.fileName.front() == '/')
+        return;
+
+    using path = boost::filesystem::path;
+    for (const auto &evt : pDef->events) {
+        if (&keyEvt != &evt
+                && path(keyEvt.fileName).filename() == path(keyEvt.fileName).filename()
```
kdudka, line 657: The expression `path(keyEvt.fileName).filename() == path(keyEvt.fileName).filename()` will always be true.  I think this should be written in a more readable way.

kdudka (changes requested):

src/lib/parser-gcc.cc
```diff
@@ -639,8 +640,30 @@ void GccPostProcessor::Private::transUbsan(Defect *pDef) const
...
+
+    // UBSAN may only use the base name for the file path in its key event.
+    // However, the top of the backtrace, if present, always contains the
+    // whole path to this file.
+    if (!keyEvt.fileName.empty() && keyEvt.fileName.front() == '/')
+        return;
+
+    using path = boost::filesystem::path;
+    for (const auto &evt : pDef->events) {
+        const auto &keyFileName = path(keyEvt.fileName).filename();
```
kdudka, line 656: I would move this assignment before the `for` loop.

src/lib/parser-gcc.cc
```diff
@@ -639,8 +640,30 @@ void GccPostProcessor::Private::transUbsan(Defect *pDef) const
...
+        const auto &evtFileName = path(evt.fileName).filename();
+
+        // check that the basenames and line numbers match
+        if (&keyEvt != &evt
+                && keyFileName == evtFileName
+                && keyEvt.line == evt.line) {
```
kdudka, line 662: I would put a newline between `)` and `{`.

kdudka (approved):

src/lib/parser-gcc.cc
lzaoral: Addressed in  be20b75d9e60afe8bb3820d668750fceadd3907b.

PR #142, opened 2023-09-22
parser-json-cov: treat defects with high impact as important

kdudka:
Resolves: https://issues.redhat.com/browse/OSH-289

### DIFF
--- a/src/lib/parser-json-cov.cc
+++ b/src/lib/parser-json-cov.cc
@@ -50,10 +50,16 @@ bool CovTreeDecoder::readNode(Defect *def)
     // out of the supported tools, only Coverity produces this data format
     def->tool = "coverity";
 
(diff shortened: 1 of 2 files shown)


hanchuntao:
I don't think anything needs to be changed, check `impact`, if it is `High`, set `imp=1`.
The only possible issue is whether all additions are complete in the test data.
If both are added， LGTM

hanchuntao (approved):
I don't think anything needs to be changed, check impact, if it is High, set imp=1.
The only possible issue is whether all additions are complete in the test data.
If all are added， LGTM

kdudka:

Note that if the test data was incorrect, the upstream CI would be red.  We run all csdiff tests in the CI matrix on Ubuntu LTS as well as all supported Fedora/EPEL releases.

PR #145, opened 2023-10-07
csgrep --warning-rate-limit: fix key events out of range

kdudka:
Related: https://issues.redhat.com/browse/OSH-67


kdudka:
Just for the context, this bug did not manifest earlier because we were using `--warning-rate-limit` only for Gitleaks, where the first event is always a key event.

hanchuntao:
After reading the context from https://issues.redhat.com/browse/OSH-67, The code looks fine, LGTM

hanchuntao (approved):

kdudka:

PR #153, opened 2023-12-12
cmake: make FindPython use system Python 3 version

kdudka:
The newest installed version of Python 3 is often not the default version on a Gentoo Linux system, where multiple Python 3 versions can be installed on a single system and /usr/bin/python3 is the one that should be used by default.

Related: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8287

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -91,6 +91,10 @@ option(PYCSDIFF_PYTHON2 "Set to ON to build pycsdiff for Python 2" OFF)
 option(PYCSDIFF_PYTHON3 "Set to ON to build pycsdiff for Python 3" ON)
 
 macro(build_pycsdiff version)
+    # make FindPython use system Python version
+    # https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8287
+    set(Python${version}_FIND_UNVERSIONED_NAMES FIRST)
+
     # check for Python libs (e.g. python${version}-devel on Fedora)
     # Interpreter is required for Python_SITEARCH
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -91,6 +91,10 @@ option(PYCSDIFF_PYTHON2 "Set to ON to build pycsdiff for Python 2" OFF)
 option(PYCSDIFF_PYTHON3 "Set to ON to build pycsdiff for Python 3" ON)
 
 macro(build_pycsdiff version)
+    # make FindPython use system Python 3 version
+    # https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8287
+    set(Python3_FIND_UNVERSIONED_NAMES FIRST)
```
lzaoral, line 96: Make the variable generic because `FindPython2` supports `Python2_FIND_UNVERSIONED_NAMES` as well:
```suggestion
    # make FindPython${version} use system Python version
    # https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8287
    set(Python${version}_FIND_UNVERSIONED_NAMES FIRST)
```

lzaoral (changes requested):

src/CMakeLists.txt
kdudka: Addressed in 133076526aa324eee6748d272a6c39d110312b7f.

lzaoral (approved):

kdudka:
@lzaoral Thanks for review!

PR #159, opened 2024-01-31
parser-common: recognize events produced by SIGMA checkers

kdudka:


src/lib/parser-common.hh
```diff
@@ -35,7 +35,8 @@
 
 #define RE_EVENT_GCC "(?:(?:(?:fatal|internal|runtime) )?[A-Za-z][A-Za-z0-9_-]+)(?:\\[[^ \\]]+\\])?"
 #define RE_EVENT_PROSPECTOR "(?:[A-Z]+[0-9]+\\[[a-z0-9-]+\\])"
-#define RE_EVENT RE_EVENT_GCC "|" RE_EVENT_PROSPECTOR
+#define RE_EVENT_SIGMA "(?:Sigma (?:main )?event)"
```
hanchuntao, line 38: Should we need to define `remediation `?

kdudka: `remediation` is already covered by `RE_EVENT_GCC` as all other events that do not contain spaces or any special delimiters.  This event specifically had already been also covered by a few tests.
```
% grep remediation: -r tests
```

kdudka:
The failing CI job seems unrelated to the changes in this pull request.  I have created #160 for it.

src/lib/parser-common.hh
hanchuntao: ok, got it

hanchuntao (approved):

Issue #160, opened 2024-01-31
different sarif output produced on macOS latest


kdudka:
The `macOS latest` CI job started to fail due to differences in the SARIF output produced in this environment:
```diff
--- /Users/runner/work/csdiff/csdiff/tests/csgrep/0080-sarif-writer-stdout.txt	2024-01-31 11:26:22.000000000 +0000
+++ -	2024-01-31 11:33:56.000000000 +0000
@@ -1,13 +1,11 @@
 {
     "$schema": "https://json.schemastore.org/sarif-2.1.0.json",
     "version": "",
-    "inlineExternalProperties": [
-        {
-            "externalizedProperties": {
-                "analyzer-version-snyk-code": "1.0.0"
-            }
+    "inlineExternalProperties": {
+        "externalizedProperties": {
+            "analyzer-version-snyk-code": "1.0.0"
         }
-    ],
+    },
     "runs": [
         {
             "tool": {
```
affected CI job: https://github.com/csutils/csdiff/actions/runs/7725324164/job/21059295165?pr=159

lzaoral:
The issue was caused by the following change introduced in Boost JSOn 1.84.0: https://github.com/boostorg/json/pull/921

PR #161, opened 2024-01-31
writer-json-sarif: fix miscompilation with recent Apple Clang

lzaoral:
For a reason I'm too lazy to diagnose, latest releases of Apple Clang now ignore the outermost initializer lists with object initializers which GCC still translates to an array.  Construct the arrays explicitly to avoid this ambiguity.

Resolves: https://github.com/csutils/csdiff/issues/160


kdudka (approved):
@lzaoral Wow, that was fast.  Thanks!

PR #162, opened 2024-02-06
json-parser: recognize SARIF format for semgrep output

rhyw:
Resolves: https://issues.redhat.com/browse/OSH-57
Closes: https://github.com/csutils/csdiff/pull/162


tests/csgrep/0120-sarif-parser-semgrep-args.txt
```diff
@@ -0,0 +1 @@
+--mode=json
```
hanchuntao, line 1: add a empty line at the end of file

tests/csgrep/CMakeLists.txt
```diff
@@ -163,3 +163,4 @@ test_csgrep("0116-csgrep-warning-rate-limit"          )
 test_csgrep("0117-csgrep-set-imp-level"               )
 test_csgrep("0118-gcc-parser-ubsan-dedup"             )
 test_csgrep("0119-cov-parser-sigma"                   )
+test_csgrep("0120-sarif-parser-semgrep"               )
```
hanchuntao, line 166: add a empty line at the end of file

tests/csgrep/0120-sarif-parser-semgrep-args.txt
rhyw: I think no empty line should be added, this file will be read be the test script. the other reason is there're lots of similar `*-args.txt` under this directory, we better keep it as it is.

tests/csgrep/CMakeLists.txt
rhyw: I don't have a very good understanding of this `CMakeList.txt`, I better not add it since originally there's no.

src/lib/parser-json-sarif.cc
```diff
@@ -130,6 +130,13 @@ void SarifTreeDecoder::readScanProps(
         if (!version.empty())
             (*pDst)["analyzer-version-gitleaks"] = std::move(version);
     }
+    else if (name == "Semgrep OSS") {
```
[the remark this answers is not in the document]
rhyw: this is not decided by us, it's the semgrep cli tool produced this key.

tests/csgrep/0120-sarif-parser-semgrep-stdout.txt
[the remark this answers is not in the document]
rhyw: My attempt was very tricky - I don't find from anywhere the helpers on generating files like this.

What I did was to use the plugin `semgrep` to scan a source rpm, so I get a `.sarif` file and a `.js`(json) file. I then tweak the sarif file as stdin and the `scan-results-all.js` as stdout. I did some investigation and guess that's how these text files were feeded by other plugins(snyk, gitleaks for example). If I understand it correctly, it's used by `csgrep`(used in the `filter_hook`) to test if after processing the output is as expected.

By tweaking the sarif file, I mean the original scan may contain tens of thousands of lines, thus I need to remove some redundant lines, for example I randomly chose a `gzip` source build, there're 700+ semgrep rules claimings with approx 70k+ lines, we definitely don't want to include all of them. Other tweaks includes for example to alter some of the lines produced, for example the unintended suspicous temporary dir prefix in the rules path. I think the process is quite time consuming without a guide. Also I doubt if there's a better way of doing this.

cc @kdudka

tests/csgrep/0120-sarif-parser-semgrep-args.txt
kdudka: @hanchuntao Did you rather mean a new line?  I am not sure if the PR was updated but I can see a new line at the end of the file:
```sh
% hexdump -C tests/csgrep/0120-sarif-parser-semgrep-args.txt
00000000  2d 2d 6d 6f 64 65 3d 6a  73 6f 6e 0a              |--mode=json.|
0000000c
```

tests/csgrep/0120-sarif-parser-semgrep-stdout.txt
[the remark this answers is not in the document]
kdudka: @hanchuntao There is a script to sync the expected output of tests: https://github.com/csutils/csdiff/blob/main/tests/csgrep/sync.sh

tests/csgrep/CMakeLists.txt
kdudka: There **is** a new line at the end of the file.  An empty line is not needed.

kdudka:
@rhyw OSH-57 should be referred to as `Related: ` rather than `Resolves: ` because this pull request does not resolve the issue on its own.

lzaoral (approved):
LGTM

rhyw:
> @rhyw OSH-57 should be referred to as `Related: ` rather than `Resolves: ` because this pull request does not resolve the issue on its own.

Thanks. Commit message / PR description updated.

kdudka (approved):
@rhyw Looks good.  Thanks for the update!

PR #164, opened 2024-02-19
csgrep: Fix --prune-events option description

thoger:
Correct description of the --prune-events option for csgrep.  When this option is specified, events with the verbosity level matching the number specified as option argument are preserved, while the previous description indicated they were pruned.

Also slightly tweaking the --imp-level description to replace "greater or equal than" with "greater than or equal to", similar to the wording of the new description for --prune-events.


kdudka (approved):
Makes perfect sense.  Thanks for the contribution!

Issue #136, opened 2023-08-15
Please include `region` data in SARIF if available :vertical_traffic_light: 


jamacku:
In the current implementation, only `region.startLine` is populated. I would like to see also `endColumn`, `endLine`, and `startColumn`.

#### Example

```sh
shellcheck --format=json1 src/index.sh | csgrep --mode=sarif
```

ShellCheck output:

```json
{
  "comments": [
    {
      "file": "src/index.sh",
      "line": 4,
      "endLine": 4,
      "column": 27,
      "endColumn": 43,
      "level": "info",
      "code": 2312,
      "message": "Consider invoking this command separately to avoid masking its return value (or use '|| true' to ignore).",
      "fix": null
    },
  ],
}
```

csgrep output:

```json
{
    "$schema": "https://json.schemastore.org/sarif-2.1.0.json",
    "version": "2.1.0",
    "runs": [
        {
            "results": [
                {
                    "ruleId": "SHELLCHECK_WARNING: info[SC2312]",
                    "locations": [
                        {
                            "id": 0,
                            "physicalLocation": {
                                "artifactLocation": {
                                    "uri": "src/index.sh"
                                },
                                "region": {
                                    "startLine": 4
                                }
                            }
                        }
                    ],
                    "message": {
                        "text": "Consider invoking this command separately to avoid masking its return value (or use '|| true' to ignore)."
                    },
                },
            ]
        }
    ]
}
```

I would like to see:

```json
{
    "startLine": 4
    "startColumn": 27,
    "endLine": 4,
    "endColumn": 43,
}
```

It could greatly improve user experience on platforms like GitHub that can display SARIF data in UI.

kdudka:
@jamacku Sorry for the delay!  I have submitted a pull request to implement this: https://github.com/csutils/csdiff/pull/167

jamacku:
Thank you @kdudka 🚀

PR #167, opened 2024-03-29
propagate endLine/endColumn

kdudka:
Initially csdiff supported only `startLine`/`startColumn`.  Some scanners provide also `endLine`/`endColumn` to denote a region of certain size.

The extra information is stored efficiently in the internal data structure and in csdiff's native JSON format by tracking the diff between `startLine`/`endLine` and `startColumn`/`endColumn`, respectively, and only when the diff is a non-zero (positive) number.

Resolves: https://github.com/csutils/csdiff/issues/136

### DIFF
--- a/src/lib/defect.hh
+++ b/src/lib/defect.hh
@@ -20,6 +20,7 @@
 #ifndef H_GUARD_DEFECT_H
 #define H_GUARD_DEFECT_H
 
+#include <limits>
 #include <map>
 #include <string>
 #include <vector>
@@ -54,6 +55,10 @@ struct DefEvent {
     /// 0 = key event,  1 = info event,  2 = trace event
     int                 verbosityLevel  = 0;
 
+    using TNumDiff = unsigned short;
+    TNumDiff            hSize           = 0; //< endColumn - startColumn
+    TNumDiff            vSize           = 0; //< endLine - startLine
+
     DefEvent() { }
 
     explicit DefEvent(const std::string &event):
@@ -61,6 +66,15 @@ struct DefEvent {
     {
     }
 };
+
+///< return (end - beg) if it is positive and fits into target type, 0 otherwise
+inline DefEvent::TNumDiff diffNums(const int beg, const int end)
+{
+    const int diff = end - beg;
+    return (0 < diff && diff < std::numeric_limits<DefEvent::TNumDiff>::max())
+        ? diff
+        : 0;
+}
 
 inline bool cmpEvents(bool *pResult, const DefEvent &a, const DefEvent &b)
 {
@@ -70,6 +84,8 @@ inline bool cmpEvents(bool *pResult, const DefEvent &a, const DefEvent &b)
     RETURN_BY_REF_IF_COMPARED(a, b, event);
     RETURN_BY_REF_IF_COMPARED(a, b, msg);
     RETURN_BY_REF_IF_COMPARED(a, b, verbosityLevel);
+    RETURN_BY_REF_IF_COMPARED(a, b, hSize);
+    RETURN_BY_REF_IF_COMPARED(a, b, vSize);
 
     // incomparable events
     return false;
--- a/src/lib/parser-json-gcc.cc
+++ b/src/lib/parser-json-gcc.cc
@@ -21,6 +21,8 @@
 
 #include "parser-gcc.hh"            // for GccPostProcessor
 
+using std::string;
+
 struct GccTreeDecoder::Private {
     GccPostProcessor            postProc;
 };
@@ -32,10 +34,56 @@ GccTreeDecoder::GccTreeDecoder():
 
 GccTreeDecoder::~GccTreeDecoder() = default;
 
+static bool gccReadLocRegion(
+        DefEvent           *pEvt,
+        const pt::ptree    &start,
+        const pt::ptree    &finish)
+{
+    // read file name
+    pEvt->fileName = valueOf<string>(start, "file", "<unknown>");
+    if (pEvt->fileName != valueOf<string>(finish, "file", "<unknown>"))
+        return false;
+
+    // read line
+    if ((pEvt->line = valueOf<int>(start, "line"))) {
+        const int endLine = valueOf<int>(finish, "line");
+        pEvt->vSize = diffNums(pEvt->line, endLine);
+    }
+
+    // read column
+    if ((pEvt->column = valueOf<int>(start, "byte-column"))) {
+        const int endColumn = valueOf<int>(finish, "byte-column");
+        pEvt->hSize = diffNums(pEvt->column, endColumn);
+    }
+
+    return true;
+}
+
+static void gccReadLocation(DefEvent *pEvt, const pt::ptree *locs)
+{
+    if (locs->empty())
+        return;
+
+    const pt::ptree &firstLoc = locs->begin()->second;
+
+    // try to read a region between start..finish
+    const pt::ptree *start, *finish;
+    if (findChildOf(&start, firstLoc, "start")
+            && findChildOf(&finish, firstLoc, "finish")
+            && gccReadLocRegion(pEvt, *start, *finish))
+        return;
+
+    // fallback to caret
+    const pt::ptree *caret;
+    if (findChildOf(&caret, firstLoc, "caret")) {
+        pEvt->fileName  = valueOf<string>(*caret, "file", "<unknown>");
+        pEvt->line      = valueOf<int>   (*caret, "line");
+        pEvt->column    = valueOf<int>   (*caret, "byte-column");
+    }
+}
+
 static bool gccReadEvent(DefEvent *pEvt, const pt::ptree &evtNode)
 {
-    using std::string;
-
     // read kind (error, warning, note)
     string &evtName = pEvt->event;
     evtName = valueOf<string>(evtNode, "kind");
@@ -45,14 +93,8 @@ static bool gccReadEvent(DefEvent *pEvt, const pt::ptree &evtNode)
     // read location
     pEvt->fileName = "<unknown>";
     const pt::ptree *locs;
-    if (findChildOf(&locs, evtNode, "locations") && !locs->empty()) {
-        const pt::ptree *caret;
-        if (findChildOf(&caret, locs->begin()->second, "caret")) {
-            pEvt->fileName  = valueOf<string>(*caret, "file", "<unknown>");
-            pEvt->line      = valueOf<int>   (*caret, "line");
-            pEvt->column    = valueOf<int>   (*caret, "byte-column");
-        }
-    }
+    if (findChildOf(&locs, evtNode, "locations"))
+        gccReadLocation(pEvt, locs);
 
     // read message
     pEvt->msg = valueOf<string>(evtNode, "message", "<unknown>");
--- a/src/lib/parser-json-simple.cc
+++ b/src/lib/parser-json-simple.cc
@@ -71,8 +71,10 @@ SimpleTreeDecoder::Private::Private(InStream &input):
         "column",
         "event",
         "file_name",
+        "h_size",
         "line",
         "message",
+        "v_size",
         "verbosity_level",
     };
 }
@@ -134,6 +136,8 @@ bool SimpleTreeDecoder::readNode(Defect *def)
     TEvtList &evtListDst = def->events;
     const pt::ptree &evtListSrc = defNode.get_child("events");
     for (const pt::ptree::value_type &evtItem : evtListSrc) {
+        using TNumDiff = DefEvent::TNumDiff;
+
         const pt::ptree &evtNode = evtItem.second;
         d->reportUnknownNodes(Private::NK_EVENT, evtNode);
 
@@ -141,6 +145,8 @@ bool SimpleTreeDecoder::readNode(Defect *def)
         evt.fileName    = valueOf<std::string   >(evtNode, "file_name");
         evt.line        = valueOf<int           >(evtNode, "line");
         evt.column      = valueOf<int           >(evtNode, "column");
+        evt.hSize       = valueOf<TNumDiff      >(evtNode, "h_size");
+        evt.vSize       = valueOf<TNumDiff      >(evtNode, "v_size");
         evt.event       = valueOf<std::string   >(evtNode, "event");
         evt.msg         = valueOf<std::string   >(evtNode, "message");
         evt.verbosityLevel = valueOf<int>(evtNode, "verbosity_level", -1);
--- a/tests/csgrep/0090-sarif-writer-illegal-utf8-sequence-stdout.txt
+++ b/tests/csgrep/0090-sarif-writer-illegal-utf8-sequence-stdout.txt
@@ -30,9 +30,6 @@
                             "physicalLocation": {
                                 "artifactLocation": {
                                     "uri": "test.c"
-                                },
-                                "region": {
-                                    "startLine": 0
                                 }
                             }
                         }
@@ -51,9 +48,6 @@
                                                 "physicalLocation": {
                                                     "artifactLocation": {
                                                         "uri": "test.c"
-                                                    },
-                                                    "region": {
-                                                        "startLine": 0
                                                     }
                                                 },
                                                 "message": {
--- a/tests/csgrep/0094-gcc-json-stdout.txt
+++ b/tests/csgrep/0094-gcc-json-stdout.txt
@@ -49,7 +49,8 @@
                 {
                     "file_name": "xxx.c",
                     "line": 11,
-                    "column": 15,
+                    "column": 11,
+                    "h_size": 7,
                     "event": "warning[-Wunused-parameter]",
                     "message": "unused parameter ‘argc’",
                     "verbosity_level": 0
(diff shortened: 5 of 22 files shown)


src/lib/defect.hh
```diff
@@ -54,6 +55,9 @@ struct DefEvent {
     /// 0 = key event,  1 = info event,  2 = trace event
     int                 verbosityLevel  = 0;
 
+    short               hSize           = 0; //< endColumn - startColumn
+    short               vSize           = 0; //< endLine - startLine
```
lzaoral, line 59: I'd probably use `unsigned short`, because `diffNums` already makes sure that the stored value is not negative.

src/lib/defect.hh
```diff
@@ -62,6 +66,15 @@ struct DefEvent {
     }
 };
 
+///< return (end - beg) if it is positive and fits into `short`, 0 otherwise
+inline short diffNums(const int beg, const int end)
+{
+    const int diff = end - beg;
+    return (0 < diff && diff < std::numeric_limits<short>::max())
+        ? diff
+        : 0;
```
lzaoral, line 75: OT: When we switch to C++17, we can repalce this with `std::clamp`: https://en.cppreference.com/w/cpp/algorithm/clamp

src/lib/parser-json-gcc.cc
```diff
@@ -32,6 +32,56 @@ GccTreeDecoder::GccTreeDecoder():
 
 GccTreeDecoder::~GccTreeDecoder() = default;
 
+static bool gccReadLocRegion(
+        DefEvent           *pEvt,
+        const pt::ptree    &beg,
+        const pt::ptree    &end)
+{
+    // read file name
```
lzaoral, line 40: Maybe add `using std::string` for consistency with the functions below?

src/lib/parser-json-gcc.cc
```diff
@@ -32,6 +32,56 @@ GccTreeDecoder::GccTreeDecoder():
...
+    if (locs->empty())
+        return;
+
+    const pt::ptree &firstLoc = locs->begin()->second;
+
+    // try to read a region between start..finish
+    const pt::ptree *beg, *end;
+    if (findChildOf(&beg, firstLoc, "start")
+            && findChildOf(&end, firstLoc, "finish")
+            && gccReadLocRegion(pEvt, *beg, *end))
```
lzaoral, line 73: Nit:  I'd use different names for the pointers because the first thing that came to my mind I saw `*end` was:
> Why is there a deference of end iterator?

(The same remark applies also to the function above.)

src/lib/parser-json-gcc.cc
```diff
@@ -32,6 +32,56 @@ GccTreeDecoder::GccTreeDecoder():
...
+        return false;
+
+    // read line
+    if ((pEvt->line = valueOf<int>(beg, "line"))) {
+        const int endLine = valueOf<int>(end, "line");
+        pEvt->vSize = diffNums(pEvt->line, endLine);
+    }
+
+    // read column
+    if ((pEvt->column = valueOf<int>(beg, "byte-column"))) {
```
lzaoral, line 52: Is it possible that GCC will generate report that will have `0` for the start line and some positive value for the finish line?  (The same goes for columns.)

jamacku (commented):
Thank you for working on this RFE. I have tried to test it, and unfortunately, the `endLine`/`endColumn` properties aren't filled for the following shell script.

```
[root@vm-10-0-187-125 ~]# cat main.sh 
# shell.sh
echo $1                           # Unquoted variables
find . -name *.ogg                # Unquoted find/grep patterns
rm "~/my file.txt"                # Quoted tilde expansion
v='--verbose="true"'; cmd $v      # Literal quotes in variables
for f in "*.ogg"                  # Incorrectly quoted 'for' loops
touch $@                          # Unquoted $@

[root@vm-10-0-187-125 ~]# shellcheck --format=json1 main.sh | csgrep --mode=sarif  | sarif-fmt
error: Couldn't parse this for loop. Fix to allow more checks.
  ┌─ main.sh:6:1
  │
6 │ for f in "*.ogg"                  # Incorrectly quoted 'for' loops
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: Expected 'do'.
  ┌─ main.sh:7:1
  │
7 │ touch $@                          # Unquoted $@
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: Expected 'do'. Fix any mentioned problems and try again.
  ┌─ main.sh:7:1
  │
7 │ touch $@                          # Unquoted $@
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: 0 errors emitted

[root@vm-10-0-187-125 ~]# shellcheck --format=json main.sh | shellcheck-sarif  | sarif-fmt
error: Couldn't parse this for loop. Fix to allow more checks.
  ┌─ main.sh:6:1
  │
6 │ for f in "*.ogg"                  # Incorrectly quoted 'for' loops
  │ ^
  │
  = SC1073
  = For more information: https://www.shellcheck.net/wiki/SC1073

error: Expected 'do'.
  ┌─ main.sh:7:1
  │
7 │ touch $@                          # Unquoted $@
  │ ^
  │
  = SC1058
  = For more information: https://www.shellcheck.net/wiki/SC1058

error: Expected 'do'. Fix any mentioned problems and try again.
  ┌─ main.sh:7:1
  │
7 │ touch $@                          # Unquoted $@
  │ ^
  │
  = SC1072
  = For more information: https://www.shellcheck.net/wiki/SC1072

error: 0 errors emitted
```

But when I commented last two lines, so ShellCheck could parse the whole file, the properties are set correctly by `csgrep`

```
[root@vm-10-0-187-125 ~]# cat main.sh 
# shell.sh
echo $1                           # Unquoted variables
find . -name *.ogg                # Unquoted find/grep patterns
rm "~/my file.txt"                # Quoted tilde expansion
v='--verbose="true"'; cmd $v      # Literal quotes in variables
#for f in "*.ogg"                  # Incorrectly quoted 'for' loops
#touch $@                          # Unquoted $@

shellcheck --format=json1 main.sh | csgrep --mode=sarif  | sarif-fmt
error: Tips depend on target shell and yours is unknown. Add a shebang or a 'shell' directive.
  ┌─ main.sh:1:1
  │
1 │ # shell.sh
  │ ^^^^^^^^^^

warning: Double quote to prevent globbing and word splitting.
  ┌─ main.sh:2:6
  │
2 │ echo $1                           # Unquoted variables
  │      ^^

warning: Quote the parameter to -name so the shell won't interpret it.
  ┌─ main.sh:3:14
  │
3 │ find . -name *.ogg                # Unquoted find/grep patterns
  │              ^^^^^

warning: Use ./*glob* or -- *glob* so names with dashes won't become options.
  ┌─ main.sh:3:14
  │
3 │ find . -name *.ogg                # Unquoted find/grep patterns
  │              ^

warning: Tilde does not expand in quotes. Use $HOME.
  ┌─ main.sh:4:5
  │
4 │ rm "~/my file.txt"                # Quoted tilde expansion
  │     ^^^^^^^^^^^^^

warning: Quotes/backslashes will be treated literally. Use an array.
  ┌─ main.sh:5:3
  │
5 │ v='--verbose="true"'; cmd $v      # Literal quotes in variables
  │   ^^^^^^^^^^^^^^^^^^

warning: Quotes/backslashes in this variable will not be respected.
  ┌─ main.sh:5:27
  │
5 │ v='--verbose="true"'; cmd $v      # Literal quotes in variables
  │                           ^^

warning: 6 warnings emitted
error: 6 errors emitted
```

src/lib/parser-json-gcc.cc
lzaoral: And the same also goes for different input formats.

src/lib/parser-json-simple.cc
```diff
@@ -141,6 +143,8 @@ bool SimpleTreeDecoder::readNode(Defect *def)
         evt.fileName    = valueOf<std::string   >(evtNode, "file_name");
         evt.line        = valueOf<int           >(evtNode, "line");
         evt.column      = valueOf<int           >(evtNode, "column");
+        evt.hSize       = valueOf<short         >(evtNode, "h_size");
+        evt.vSize       = valueOf<short         >(evtNode, "v_size");
```
lzaoral, line 147: Do not forget to update these as well, if you change the type to `unsigned short`.

tests/csgrep/0094-gcc-json-stdout.txt
```diff
@@ -49,7 +49,8 @@
                 {
                     "file_name": "xxx.c",
                     "line": 11,
-                    "column": 15,
+                    "column": 11,
+                    "h_size": 7,
```
lzaoral, line 53: This is suspicious.  Other updated test outputs did not change existing fields.

tests/csgrep/0094-gcc-json-stdout.txt
```diff
@@ -65,7 +66,8 @@
                 {
                     "file_name": "xxx.c",
                     "line": 11,
-                    "column": 27,
+                    "column": 21,
+                    "h_size": 11,
```
lzaoral, line 70: Ditto.

lzaoral (changes requested):

kdudka:
@jamacku Thank you for testing it!  The problem you pointed out should be fixed in b67c44fbc8cbbf180f051105bf341c814fb3102a.

src/lib/defect.hh
kdudka: The idea was to avoid mixing signed/unsigned types while adding them to `line` and `column`, which are signed.  On the other hand `usigned short` allows us to propagate one more bit.  I have also created the `TNumDiff` type name so that this is easier to change later on.  Committed as 10afff8d6350174311c1e79555aea9143ff9c88a.

src/lib/defect.hh
kdudka: `std::clamp` with return `USHORT_MAX` for numbers above the limit whereas `diffNums()` returns 0 in this case, which means the value is dropped.

src/lib/parser-json-gcc.cc
kdudka: I have moved it to the global level (after includes) to simplify it.  Fixed in e886689876a46b9d682b10c8d9fe6f4de44ab770.

src/lib/parser-json-gcc.cc
kdudka: Renamed to `start` and `finish` in e886689876a46b9d682b10c8d9fe6f4de44ab770.

src/lib/parser-json-gcc.cc
kdudka: All the tools I know number lines starting with 1.  If line number is 0, it means that line number is not available for the finding.  In that case, there is no region to size with `h_size`/`v_size`.

src/lib/parser-json-simple.cc
kdudka: Fixed in 10afff8d.

jamacku (commented):
Works great, Thank you

```
[root@vm-10-0-187-125 ~]# shellcheck --format=json1 main.sh | csgrep --mode=sarif  | sarif-fmt
error: Couldn't parse this for loop. Fix to allow more checks.
  ┌─ main.sh:6:1
  │
6 │ for f in "*.ogg"                  # Incorrectly quoted 'for' loops
  │ ^

error: Expected 'do'.
  ┌─ main.sh:7:1
  │
7 │ touch $@                          # Unquoted $@
  │ ^

error: Expected 'do'. Fix any mentioned problems and try again.
  ┌─ main.sh:7:1
  │
7 │ touch $@                          # Unquoted $@
  │ ^

error: 0 errors emitted
```
```
[root@vm-10-0-187-125 ~]# shellcheck --format=json1 main.sh | csgrep --mode=sarif  | sarif-fmt
error: Tips depend on target shell and yours is unknown. Add a shebang or a 'shell' directive.
  ┌─ main.sh:1:1
  │
1 │ # shell.sh
  │ ^

warning: Double quote to prevent globbing and word splitting.
  ┌─ main.sh:2:6
  │
2 │ echo $1                           # Unquoted variables
  │      ^^

warning: Quote the parameter to -name so the shell won't interpret it.
  ┌─ main.sh:3:14
  │
3 │ find . -name *.ogg                # Unquoted find/grep patterns
  │              ^^^^^

warning: Use ./*glob* or -- *glob* so names with dashes won't become options.
  ┌─ main.sh:3:14
  │
3 │ find . -name *.ogg                # Unquoted find/grep patterns
  │              ^

warning: Tilde does not expand in quotes. Use $HOME.
  ┌─ main.sh:4:5
  │
4 │ rm "~/my file.txt"                # Quoted tilde expansion
  │     ^^^^^^^^^^^^^

warning: Quotes/backslashes will be treated literally. Use an array.
  ┌─ main.sh:5:3
  │
5 │ v='--verbose="true"'; cmd $v      # Literal quotes in variables
  │   ^^^^^^^^^^^^^^^^^^

warning: Quotes/backslashes in this variable will not be respected.
  ┌─ main.sh:5:27
  │
5 │ v='--verbose="true"'; cmd $v      # Literal quotes in variables
  │                           ^^

warning: 6 warnings emitted
error: 6 errors emitted
```

tests/csgrep/0094-gcc-json-stdout.txt
kdudka: The native JSON output of GCC is too rich for `csdiff` to handle it (and obsoleted by GCC's SARIF output already).  It provides up to three pairs of line/column:

- `start`
- `finish`
- `caret` (may point somewhere between `start` and `finish`)

The initial implementation of `GccTreeDecoder` completely ignored `start` a `finish` because `caret` was the only object that was available in all events (except events that had no location info at all).

This pull request changes `GccTreeDecoder` to use `start`/`finish` in a way that is compatible with SARIF's region object if both `start` and `finish` are available.  If any of them is missing (you can find examples in csdiff's test-suite), there is a fallback to `caret`.

Does it make sense?

tests/csgrep/0094-gcc-json-stdout.txt
kdudka: Ditto.

tests/csgrep/0094-gcc-json-stdout.txt
lzaoral: Yes, it makes perfect sense.  Thank you for the explanation, @kdudka!

kdudka:
@jamacku Perfect.  Thank you for testing it!

tests/csgrep/0090-sarif-writer-illegal-utf8-sequence-stdout.txt
```diff
@@ -32,7 +32,8 @@
                                     "uri": "test.c"
                                 },
                                 "region": {
-                                    "startLine": 0
+                                    "startLine": 0,
+                                    "endLine": 0
```
lzaoral, line 36: nit: The input file is an artificial test created in 620dc470d46908c64d3a5d14c78e015c71836aee which does not include line information at all.  Shall we fix that?

tests/csgrep/0094-gcc-json-stdout.txt
lzaoral: Oh, now I get it why it is called `caret`:
```
test.c: In function ‘main’:
test.c:4:13: warning: ‘a’ is used uninitialized [-Wuninitialized]
    4 |     return a[0];
      |            ~^~~
```
It is the location of the literal `^` caret character in the diagnostic and `start` and `finish` seem to correspond to the boundaries of the `~` squiggly line.

lzaoral:
@jamacku This is probably unrelated to this PR but the first output from `sarif-fmt` in https://github.com/csutils/csdiff/pull/167#pullrequestreview-1976532916 is weird.  There three errors but the summary shows:
```
error: 0 errors emitted
```

tests/csgrep/0090-sarif-writer-illegal-utf8-sequence-stdout.txt
jamacku: It might be good to have such data for test purposes.

jamacku:
> @jamacku This is probably unrelated to this PR but the first output from `sarif-fmt` in [#167 (review)](https://github.com/csutils/csdiff/pull/167#pullrequestreview-1976532916) is weird. There three errors but the summary shows:
> 
> ```
> error: 0 errors emitted
> ```

TBH, I'm not sure what those values should represent. It is possible that values might be a bit misleading. For example the second report shows number of errors 6 but there is 7 reports.

lzaoral (approved):
Approving.  The comments above are either minor remarks or questions.

jamacku (approved):
LGTM, works with `sarif-fmt`

tests/csgrep/0090-sarif-writer-illegal-utf8-sequence-stdout.txt
kdudka: According to https://sarifweb.azurewebsites.net/Validation the zeroes trigger validation errors.  I will replace them with 1.

kdudka: Nope.  This is not a test input but invalid output of csdiff's SARIF writer.  The correct fix is to not write the `region` object in case we have no line numbers.

kdudka: Fixed in 1d29439addac86d5fe0265f2c6405e69f29d23ad.

Issue #98, opened 2022-12-21
SARIF: Please add support for `fingerprints` in SARIF format :feet: 


jamacku:
Fingerprints could be beneficial when used with GitHub and for a general comparison of scan results.

Documentation of `fingerprints` property: [link](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317654)

GitHub SARIF Documentation: [link](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning#providing-data-to-track-code-scanning-alerts-across-runs)

kdudka:
My initial thought was to compute a hash from all the data that are matched in `DefLookup::lookup()`.  The problem is that `DefLookup` is able to count occurrences of defects with equal fields (which would result in identical hashes) whereas GitHub would see all of them as a single defect.  This would not work well when a new instance of defect with previously seen fingerprint was introduced in an update.

jamacku:
Here is scala example of creating `partialFingerprints` from [`codacy/codacy-analysis-cli`](https://github.com/codacy/codacy-analysis-cli/blob/80c8baecbfea050feb5cee184051b996cdd7fed3/cli/src/main/scala/com/codacy/analysis/cli/formatter/Sarif.scala#L168):

```scala
partialFingerprints = SarifReport.PartialFingerprints("1", generatePrimaryLocationHash(issue, fileContents)))

// ...

final case class PartialFingerprints(primaryLocationStartColumnFingerprint: String, primaryLocationLineHash: String)
```

It seems they are counting only with line numbers and with the hash of line content.

kdudka:
@jamacku The actual fingerprint computation is here: https://github.com/codacy/codacy-analysis-cli/blob/80c8baecbfea050feb5cee184051b996cdd7fed3/cli/src/main/scala/com/codacy/analysis/cli/formatter/Sarif.scala#L192

They compute MD5 hash of a string concatenated from  `fileName`, `patternId`, and `lineContentsWithoutSpaces`.

jamacku:
Here is an example of how fingerprints are handled by [`github/codeql-action`](https://github.com/github/codeql-action) - [code](https://github.com/github/codeql-action/blob/bb28e7e59e2ad6c1e5400e671795b2fa1b2fca6f/src/fingerprints.ts#L250-L309)

And also one comment how it works:

[...]

PR #168, opened 2024-03-29
sarif: initial implementation of csdiff fingerprints

kdudka:
# csdiff/v0 fingerprints

It hashes the data that csdiff uses in its matching algorithm.

From the updated tests it is obvious that these hashes already
have numerous collisions on the existing test data.

# csdiff/v1 fingerprints
    
It hashes the data that csdiff uses in its matching algorithm
and the line content without spaces.  For this fingerprint to
be computed, the results need to include the line content for 
the key event in the format produced by `csgrep --embed-context`.

Related: https://issues.redhat.com/browse/OSH-9
Resolves: https://github.com/csutils/csdiff/issues/98


jamacku:
Great example when codeql fingerprints work:
- https://github.com/redhat-plumbers-in-action/differential-shellcheck/pull/376

```sh
✋ Defects, NEEDS INSPECTION
Error: SHELLCHECK_WARNING:
./src/functions.sh:309:46: warning[SC2154]: GITHUB_REF is referenced but not assigned.
#  306|   uploadSARIF () {
#  307|     is_debug && local verbose=--verbose
#  308|   
#  309|->   echo '{"commit_oid":"'"${HEAD}"'","ref":"'"${GITHUB_REF//merge/head}"'","analysis_key":"differential-shellcheck","sarif":"'"$(gzip -c output.sarif | base64 -w0)"'","tool_names":["differential-shellcheck"]}' > payload.json
#  310|   
#  311|     local curl_args=(
#  312|       "${verbose:---silent}"

Error: SHELLCHECK_WARNING:
./src/functions.sh:331:24: info[SC2312]: Consider invoking this command separately to avoid masking its return value (or use '|| true' to ignore).
#  328|   
#  329|   get_shellcheck_version () {
#  330|     local shellcheck_version
#  331|->   shellcheck_version=$(shellcheck --version | grep -w "version:" | cut -s -d ' ' -f 2)
#  332|   
#  333|     echo "${shellcheck_version}"
#  334|   }

Error: SHELLCHECK_WARNING:
./src/functions.sh:331:47: info[SC2312]: Consider invoking this command separately to avoid masking its return value (or use '|| true' to ignore).
#  328|   
#  329|   get_shellcheck_version () {
#  330|     local shellcheck_version
#  331|->   shellcheck_version=$(shellcheck --version | grep -w "version:" | cut -s -d ' ' -f 2)
#  332|   
#  333|     echo "${shellcheck_version}"
#  334|   }
```

[The code hasn't changed](https://github.com/redhat-plumbers-in-action/differential-shellcheck/pull/376/files) for the last two defects, but they were still reported by csdiff.

I'll try to check if csdiff fingerprints would give the same results as codeql.

jamacku:
Here is the `csgrep` output from before the change and after:

- [before.json](https://github.com/csutils/csdiff/files/15132319/before.json)
- [after.json](https://github.com/csutils/csdiff/files/15132318/after.json)

You should be able to reproduce the above situation form:

- https://github.com/redhat-plumbers-in-action/differential-shellcheck/pull/376

kdudka:
@jamacku Nice!  I have added it as a regression test in 95e35d316e0ed968e59d8687fddc607692fd7b47.  It passes with ca7f32a5e1ed45b7e86da091fa6a2efe7f762b29 but not without it.

jamacku (approved):
Thank you @kdudka

src/csgrep.cc
```diff
@@ -543,7 +574,8 @@ bool chainFilters(
         return false;
     }
 
-    return chainDecoratorIntArg<ImpLevelFilter>(pEng, vm, "imp-level");
+    return chainDecoratorIntArg<ImpLevelFilter>(pEng, vm, "imp-level")
+        && chainDecoratorGeneric<FingerPrintFilter>(pEng, vm, "hash-v1");
```
lzaoral, line 578: nit: sort and align?
```suggestion
    return chainDecoratorGeneric<FingerPrintFilter>(pEng, vm, "hash-v1")
        && chainDecoratorIntArg <ImpLevelFilter>   (pEng, vm, "imp-level");
```

src/lib/finger-print.cc
```diff
@@ -0,0 +1,139 @@
...
+    return hexHashStr<sha1>(str);
+}
+
+struct FingerPrinter::Private {
+    std::string basicData;
+    std::string lineContent;
+};
+
+// separator used for hashing of data composed from multiple strings
+static const std::string sep = "\n";
```
lzaoral, line 89:
```suggestion
static constexpr std::string sep = "\n";
```

src/lib/finger-print.cc
```diff
@@ -0,0 +1,139 @@
...
+        // no line content found
+        return;
+
+    // remove all white-spaces
+    static const RE reSpace = RE("\\s+");
+    *pDst = boost::regex_replace(content, reSpace, "");
+}
+
+/// return SHA1 hash of `str` as hex-encoded string
+static std::string computeHexSHA1(const std::string &str)
```
lzaoral, line 77:
```suggestion
static inline std::string computeHexSHA1(const std::string &str)
```

src/lib/finger-print.cc
```diff
@@ -0,0 +1,139 @@
...
+    assert(fpv == FPV_CSDIFF_WITH_LINE_CONTENT);
+    if (d->lineContent.empty())
+        // no line content available
+        return "";
+
+    // return SHA1 hash from basicData AND lineContent
+    return computeHexSHA1(d->basicData + sep + d->lineContent);
+}
+
+std::string FingerPrinter::getLineContent() const
```
lzaoral, line 136:
```suggestion
inline std::string FingerPrinter::getLineContent() const
```

(If the function dodo not depend on the `Private` type, I'd have suggested to just implement this method directly in the header file.)

lzaoral (changes requested):

src/lib/finger-print.cc
kdudka: This fails to compile:
```
finger-print.cc:89:30: error: the type ‘const std::string’ {aka ‘const std::__cxx11::basic_string<char>’} of ‘constexpr’ variable ‘sep’ is not literal
   89 | static constexpr std::string sep = "\n";
      |                              ^~~
In file included from /usr/include/c++/13/string:54,
                 from /home/kdudka/git/csdiff/src/lib/defect.hh:25,
                 from /home/kdudka/git/csdiff/src/lib/finger-print.hh:23,
                 from /home/kdudka/git/csdiff/src/lib/finger-print.cc:20:
/usr/include/c++/13/bits/basic_string.h:87:11: note: ‘std::__cxx11::basic_string<char>’ is not literal because:
   87 |     class basic_string
      |           ^~~~~~~~~~~~
/usr/include/c++/13/bits/basic_string.h:87:11: note:   ‘std::__cxx11::basic_string<char>’ has a non-trivial destructor
```

lzaoral: Oh, right...  I forgot that `std::string` is not `constexpr` in pure C++11.  Sorry.

src/lib/finger-print.cc
kdudka: This breaks the linking:
```
[ 72%] Linking CXX executable csdiff
/usr/bin/ld: lib/libcs.a(deflookup.cc.o): in function `DefLookup::lookup(Defect const&)':
/home/kdudka/git/csdiff/src/lib/deflookup.cc:86: undefined reference to `FingerPrinter::getLineContent[abi:cxx11]() const'
/usr/bin/ld: lib/libcs.a(deflookup.cc.o): in function `defLookupCore':
/home/kdudka/git/csdiff/src/lib/deflookup.cc:91: undefined reference to `FingerPrinter::getLineContent[abi:cxx11]() const'
collect2: error: ld returned 1 exit status
```

src/csgrep.cc
```diff
@@ -232,6 +233,36 @@ class ImpLevelFilter: public AbstractFilter {
         }
 };
 
+class FingerPrintFilter: public AbstractFilter {
+    private:
+        const std::string hashPrefix_;
+
+    public:
+        FingerPrintFilter(AbstractWriter *agent, const std::string hashPrefix):
```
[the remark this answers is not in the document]
kdudka: Fixed in 83f4b2f18e15ae302d06d415097ed75ad2e41184.

src/csgrep.cc
kdudka: Fixed in 83f4b2f18e15ae302d06d415097ed75ad2e41184.

src/lib/finger-print.hh
```diff
@@ -0,0 +1,49 @@
...
+#define H_GUARD_FINGER_PRINT_H
+
+#include "defect.hh"
+
+#include <memory>
+
+enum EFingerPrintVer {
+    FPV_CSDIFF = 0,
+    FPV_CSDIFF_WITH_LINE_CONTENT,
+    FPV_MAX
```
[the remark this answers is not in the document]
kdudka: `FPV_MAX` should always be the last item.  I have added a comment in 6e6549dd266d0fe9a4c86778252f76ee267cefb5 to make it clear.

src/lib/hash-util.hh
```diff
@@ -0,0 +1,51 @@
...
+/// compute TEng hash of `src` and return it as hex-encoded string
+template <typename TEng, typename TSrc>
+std::string hexHashStr(const TSrc &src)
+{
+    // create hash engine and process the input
+    TEng eng;
+    eng.process_bytes(src.data(), src.size());
+
+    // export the hash as an array of unsigned int
+    // FIXME: std::remove_reference is needed on el7 (boost-1.53)
```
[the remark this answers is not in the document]
kdudka: True.  I have added a separate [commit](6e6549dd266d0fe9a4c86778252f76ee267cefb5) to drop the compatibility in case it was needed later on elsewhere.

src/lib/writer-json-sarif.cc
```diff
@@ -181,6 +182,31 @@ void SarifTreeEncoder::Private::serializeRules()
...
+        // compute a fingerprint of version `v`
+        const EFingerPrintVer fpv = static_cast<EFingerPrintVer>(v);
+        const std::string fingerPrint = fp.getHash(fpv);
+        if (fingerPrint.empty())
+            // fingerprint computation failed
+            continue;
+
+        // construct the "version" -> "value" pair
+        const std::string label = "csdiff/v" + std::to_string(v);
+        fps[label] = std::move(fingerPrint);
```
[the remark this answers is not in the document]
kdudka: Fixed in 6e6549dd266d0fe9a4c86778252f76ee267cefb5.

lzaoral:
Somehow, the comment I've just written is not here. Reposting:

src/lib/writer-json-sarif.cc
```diff
@@ -194,6 +195,31 @@ void SarifTreeEncoder::Private::serializeRules()
...
+        // compute a fingerprint of version `v`
+        const EFingerPrintVer fpv = static_cast<EFingerPrintVer>(v);
+        const std::string fingerPrint = fp.getHash(fpv);
+        if (fingerPrint.empty())
+            // fingerprint computation failed
+            continue;
+
+        // construct the "version" -> "value" pair
+        const std::string label = "csdiff/v" + std::to_string(v);
+        fps.emplace(std::move(label), std::move(fingerPrint));
```
lzaoral, line 215: @kdudka Alright, in the current state (or the previous one), using `std::move` is useless. I did not realise that both strings are `const` and `std::string` has no constructors which accept a `const` r-value reference (`const std::string &&`).

lzaoral (approved):
It's up to you if you want to use a move or copy constructor for the elements of the map.  Otherwise, LGTM.

src/lib/writer-json-sarif.cc
kdudka: Thank you for clarifying it!  I have dropped 0267593b304b7b2e2a23ddb84b0c2af5a263719b to make the code easier to read.

kdudka:
@jamacku @lzaoral Thank you both for review!

PR #169, opened 2024-04-02
Added column number in Coverity results using JSONv10

jperezdealgaba:
- Added the column number to the Coverity results using Coverity APIv10 in `parser-json-cov.cc`

- Added tests for the column property in Coverity using JSON v10 results. Two different tests cases have been added: In one of them, the column number is present in the results. In the second one, the column number is null.


kdudka (approved):
@jperezdealgaba Looks good.  Thanks!

PR #172, opened 2024-04-18
parser-cov: further tweaks to key event matching

kdudka:
Resolves: https://issues.redhat.com/browse/OSH-622


kdudka:
Confirmed to work as expected by @petr-matousek at https://issues.redhat.com/browse/OSH-622.  Merging...

PR #173, opened 2024-04-22
writer-json-sarif: propagate the "imp" flag as level

kdudka:
If the "imp" flag is set on a finding, set the SARIF level to "error", which means "Important".

Related: https://issues.redhat.com/browse/OSH-565


kdudka:
Verified by Camilo Cota at https://issues.redhat.com/browse/PSSECAUT-395.  Merging...

src/lib/writer-json-sarif.cc
```diff
@@ -338,8 +338,12 @@ void SarifTreeEncoder::appendDef(const Defect &def)
         // update tool for this rule
         d->ruleMap[ruleId].tool = def.tool;
 
-    // key event severity level
-    sarifEncodeLevel(&result, keyEvt.event);
+    if (0 < def.imp)
+        // if the "imp" flag is set, promote this finding to "Important"
+        result.emplace("level", "error");
```
hanchuntao, line 343: If exist the scenario that the `level` exists, but the value is empty?  so that cannot replace the value of `level`

kdudka: Nope.  Beside the line you commented on, `level` is only assigned in `sarifEncodeLevel()`.  For each finding, the writer either assigns `level: error` explicitly, or it calls `sarifEncodeLevel()` to assign it.

PR #176, opened 2024-05-03
fix(cshtml): `.js` -> `.json`

jamacku:
Correctly reference accepted format as JSON instead of JS.


kdudka (approved):
@jamacku Thanks!  I have only tweaked the commit message.

PR #179, opened 2024-05-10
Use HTML5 instead of XHTML 1.0 :earth_africa: 

jamacku:
> [!NOTE]
>
> UTF-8 is the default encoding in HTML5


lzaoral (approved):
LGTM! I quite sure that this is safe but let's wait for @kdudka's review as well.

kdudka (approved):
@jamacku Looks good, thanks!  I have only tweaked the commit message.

PR #180, opened 2024-05-10
Fix some HTML related warnings found by HTMLLint :heavy_check_mark: 

jamacku:


lzaoral:
Thank you @jamacku!  W3C HTML validator shows following warnings:

```
Warning: The name attribute is obsolete. Consider putting an id attribute on the nearest container instead.

[From line 7, column 37; to line 8, column 15](https://validator.w3.org/nu/#l8c15)

re-wrap;'>↩<a name='def1'></a><b
```

It should be quite easy to replace `name` with `id` as suggested in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#name

jamacku:
@lzaoral Good point, resolved in 64372ac981e435a263417542394105498be02a54

kdudka (approved):
Looks good.  Thanks!

lzaoral (approved):

Issue #174, opened 2024-04-25
Follow up on "Web interface: design/accessibility issues" issue


siteshwar:
This is a follow up on https://github.com/openscanhub/openscanhub/issues/214.

As mentioned in the [this comment](https://github.com/openscanhub/openscanhub/issues/214#issuecomment-1873817296), the report is generated through `csdiff` and should be fixed in [writer-html.cc](https://github.com/csutils/csdiff/blob/df1fda07a6fd6bb81d717d253d170935feb0886a/src/lib/writer-html.cc).

kdudka:
TL;DR The HTML writer needs to use CSS classes rather than inlining a style definition in each HTML tag separately.

lzaoral:
I suggest to keep this opened because only a subset of the issues described in https://github.com/openscanhub/openscanhub/issues/214 was resolved.  The generated HTML report still uses a colour scheme which is too hard to read for some people.

PR #181, opened 2024-05-10
use CSS instead of inlining styles :paintbrush: 

jamacku:
The branch contains additional commits from:
- #179 
- #180 

Once they are accepted, I'll drop them.

This is just a DRAFT. There is probably a better way how to implement CSS into `writer-html`

### DIFF
--- a/src/lib/writer-html.cc
+++ b/src/lib/writer-html.cc
@@ -50,10 +50,28 @@ namespace HtmlLib {
     void initHtml(std::ostream &str, std::string title) {
         escapeText(title);
 
-        str << "<!DOCTYPE html>\n\
-<html>\n\
-<head><title>" << title << "</title></head>\n\
-<body style='background: white;'>\n<h1>" << title << "</h1>\n";
+        str << R"(<!DOCTYPE html>
+<html>
+<head>
+    <title>)" << title << R"(</title>
+    <style>
+        body { background-color: white; }
+        pre { white-space: pre-wrap; }
+        .checker { background:#C0FF00; }
+        .ctxLine { color:#000000; }
+        .infoEvent { color:#808080; }
+        .infoEventComment { color:#00C0C0; }
+        .traceEvent { color: #C0C0C0; }
+        .newDefect { color: #00FF00; }
+        .parseWarning { color: #FF0000; }
+        .impFlag { color: #FF0000; font-weight: bold; }
+        #scanProps { font-family: monospace; }
+        .oddRow { background-color: #EEE; }
+        .scanPropName { padding-right: 8px; white-space: nowrap; }
+    </style>
+</head>
+<body>
+<h1>)" << title << "</h1>\n";
     }
 
     void writeLink(
@@ -74,7 +92,7 @@ namespace HtmlLib {
     }
 
     void initPre(std::ostream &str) {
-        str << "<pre style='white-space: pre-wrap;'>\n";
+        str << "<pre>\n";
     }
 
     void finalizePre(std::ostream &str) {
@@ -133,7 +151,7 @@ void writeParseWarnings(std::ostream &str, const TScanProps &props) {
         const int count = boost::lexical_cast<int>(itCount->second);
         const int ratio = boost::lexical_cast<float>(itRatio->second);
         if (ratio < parsingRatioThr)
-            str << "<p><b style='color: #FF0000;'>warning:</b> "
+            str << "<p><b class='parseWarning'>warning:</b> "
                 "low parsing ratio: " << ratio << "%</p>\n";
 
         itCount = props.find("diffbase-cov-compilation-unit-count");
@@ -149,12 +167,12 @@ void writeParseWarnings(std::ostream &str, const TScanProps &props) {
         const int baseCount = boost::lexical_cast<int>(itCount->second);
         const int baseRatio = boost::lexical_cast<float>(itRatio->second);
         if (baseRatio < parsingRatioThr && baseRatio < ratio)
-            str << "<p><b style='color: #FF0000;'>warning:</b> "
+            str << "<p><b class='parseWarning'>warning:</b> "
                 "low parsing ratio in diff base: "
                 << baseRatio << "%</p>\n";
 
         if (!count || 100 * baseCount / count < parsingOldToNewRatioThr)
-            str << "<p><b style='color: #FF0000;'>warning:</b> "
+            str << "<p><b class='parseWarning'>warning:</b> "
                 "low count of parsed units in diff base: "
                 << baseCount << "</p>\n";
     }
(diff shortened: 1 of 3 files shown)


jamacku:
Related to #174 ?

src/lib/writer-html.cc
```diff
@@ -171,16 +186,16 @@ void writeScanProps(std::ostream &str, const TScanProps &props) {
 
     HtmlLib::initSection(str, "Scan Properties");
 
-    str << "<table style='font-family: monospace;'>\n";
+    str << "<table id='scanPropTable'>\n";
```
ferdnyc, line 189: Will there **absolutely** never be more than one instance of this table on a page, under any circumstances?

If there's any way multiple instances of it can be shown on the same page, giving it a fixed ID could come back to bite you. Consider giving it a CSS class _as well as_ (or even instead of) an ID.

The table(s) and its/their contents can be styled with class-based selectors as easily as they can be with ID-based selectors.

```css
#scanPropTable tr.oddTr {
  /* Fine if there's only one scanPropTable */
}
table.scanProps tr.oddTr {
  /* Fine with multiple instances */
}
```

ferdnyc (commented):
This definitely looks like a HUGE step in the right direction, @jamacku. Thanks for looking at it!

src/lib/writer-html.cc
```diff
@@ -50,12 +50,27 @@ namespace HtmlLib {
...
+  .checker { background:#C0FF00; }\n\
+  .ctxLine { color:#000000; }\n\
+  .infoEvent { color:#808080; }\n\
+  .infoEventComment { color:#00C0C0; }\n\
+  .traceEvent { color: #C0C0C0; }\n\
+  .newDefect { color: #00FF00; }\n\
+  .parseWarning { color: #FF0000; }\n\
+  .impFlag { color: #FF0000; font-weight: bold; }\n\
+  #scanPropTable { font-family: monospace; }\n\
+  .oddTr { background-color: #EEE; }\n\
```
kdudka, line 69: I would spell this as `oddTR` or `oddRow`.

src/lib/writer-html.cc
```diff
@@ -50,12 +50,27 @@ namespace HtmlLib {
...
+  .firstTd { padding-right: 8px; white-space: nowrap; }\n\
```
kdudka, line 70: I would spell this as `scanPropKey` or `scanPropName`.

src/lib/writer-html.cc
```diff
@@ -50,12 +50,27 @@ namespace HtmlLib {
...
+  pre { white-space: pre-wrap; }\n\
+  .checker { background:#C0FF00; }\n\
+  .ctxLine { color:#000000; }\n\
+  .infoEvent { color:#808080; }\n\
+  .infoEventComment { color:#00C0C0; }\n\
+  .traceEvent { color: #C0C0C0; }\n\
+  .newDefect { color: #00FF00; }\n\
+  .parseWarning { color: #FF0000; }\n\
+  .impFlag { color: #FF0000; font-weight: bold; }\n\
+  #scanPropTable { font-family: monospace; }\n\
```
kdudka, line 68: I would spell this as `scanPropsTable`.

src/lib/writer-html.cc
kdudka: @ferdnyc Thanks for heads up!  With the current implementation, the answer is Yes.  `cshtml --scan-props-placement` takes the following values:
- top
- bottom
- none

I think that a desirable side effect of the current solution is that one can append the `#scanPropsTable` suffix to a scan results URL if they want to make the browser scroll to the Scan Properties table (without the need to use the old style `<a name='scanPropsTable/>` tag for the anchor).

kdudka (commented):
@jamacku Looks good!  A few suggestions inline...

src/lib/writer-html.cc
kdudka: ... or simply `scanProps`.

ferdnyc: +1 for not encoding structure into the CSS classes. Today it's `table#scanProps`, but tomorrow it could be `div#scanProps`.

jamacku:
@kdudka, @ferdnyc, thanks for the review. I have addressed your comments in 5f3b933edfdf47643c77df2cd638a98087e3dd9d.

src/lib/writer-html.cc
```diff
@@ -52,8 +52,27 @@ namespace HtmlLib {
...
+  .newDefect { color: #00FF00; }\n\
+  .parseWarning { color: #FF0000; }\n\
+  .impFlag { color: #FF0000; font-weight: bold; }\n\
+  #scanProps { font-family: monospace; }\n\
+  .oddRow { background-color: #EEE; }\n\
+  .scanPropName { padding-right: 8px; white-space: nowrap; }\n\
+</style>\n\
+</head>\n\
+<body>\n<h1>"
+            << title << "</h1>\n";
```
lzaoral: nit: You may consider using raw string literals:

* https://en.cppreference.com/w/cpp/language/string_literal#Raw_string_literals
* example for HTML is here: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2007/n2146.html#Markup

jamacku: Addressed in https://github.com/csutils/csdiff/pull/181/commits/796a94d21eb33c5bbf4adaeda5e9a2005925ed9f. PTAL

src/lib/writer-html.cc
```diff
@@ -544,9 +562,9 @@ void HtmlWriter::handleDef(const Defect &def)
         const bool isCtxLine = detector.isAnyCtxLine(evt);
         if (isCtxLine) {
             const char *color = (detector.isKeyCtxLine(evt))
```
kdudka, line 564: The string does not denote a color any more.  The variable should be renamed to `style` or `styleClass`.

src/lib/writer-html.cc
```diff
@@ -169,16 +187,16 @@ void writeScanProps(std::ostream &str, const TScanProps &props) {
...
     for (TScanProps::const_reference item : props) {
-        const char *trStyle = "";
+        const char *trClass = "";
         if (++i & 1)
-            trStyle = " style='background-color: #EEE;'";
+            trClass = " class='oddRow'";
 
-        const char *tdStyle0 = "padding-right: 8px; white-space: nowrap;";
-        str << "<tr" << trStyle << "><td style='" << tdStyle0 << "'>"
+        const char *tdClass0 = "scanPropName";
```
kdudka, line 198: After defining the CSS class, the `tdClass0` variable became redundant.  I think we can simply inline it:
```diff
--- a/src/lib/writer-html.cc
+++ b/src/lib/writer-html.cc
@@ -195,8 +195,7 @@ void writeScanProps(std::ostream &str, const TScanProps &props) {
         if (++i & 1)
             trClass = " class='oddRow'";

-        const char *tdClass0 = "scanPropName";
-        str << "<tr" << trClass << "><td class='" << tdClass0 << "'>"
+        str << "<tr" << trClass << "><td class='scanPropName'>"
             << item.first << "</td><td>" << item.second << "</td></tr>\n";
     }

```

kdudka (commented):
@jamacku Thanks for the update!  Two more remarks inline...

src/lib/writer-html.cc
jamacku: Good point, thanks.

kdudka (approved):
@jamacku Looks good.  Thanks for the update!

lzaoral (approved):

PR #182, opened 2024-05-21
parser: skip BOM at the beginning of input

kdudka:
... so that data format is properly recognized even when the file starts with BOM.


lzaoral (approved):

kdudka:
@lzaoral Thanks for review!

PR #183, opened 2024-05-23
msg-filter: adapt paths from llvm-17 source tree

kdudka:
... to match the llvm-19 source tree

Resolves: https://issues.redhat.com/browse/OSH-663


lzaoral (approved):

kdudka:
@lzaoral Thanks for review!

PR #184, opened 2024-05-27
parser-gcc: fix parsing of UBSAN reports with missing function names

lzaoral:


kdudka (changes requested):
The new test-case needs to be registered at https://github.com/csutils/csdiff/blob/43ae09aee53cb049da3dc9b7b87a380b19a1925e/tests/csgrep/CMakeLists.txt#L170

lzaoral:
@kdudka I have amended an existing testcase that focuses on parsing of UBSAn backtraces so that should not be necessary.

kdudka (approved):
@lzaoral I see, my bad.  Sorry for the noise.

PR #185, opened 2024-05-28
`csgrep --warning-rate-limit`: drop location from key events

kdudka:
... to ease finding deduplication.  If a checker is too noisy, a small change in the code could unexpectedly change the location (including the file path) of the `error[too-many]` key event, which would hamper the deduplication of findings.

Related: https://issues.redhat.com/browse/OSH-496


jperezdealgaba:
I can not speak for the correctness of the code, but I have been testing this for https://github.com/csutils/csdiff/pull/187 and it works as expected.

So /LGTM

lzaoral (approved):
LGTM!  Thank you, @kdudka!

kdudka:
@lzaoral Thanks for review!

PR #187, opened 2024-06-11
msg-filters: added filter for too-many event in csdiff

jperezdealgaba:
Resolves: issues.redhat.com/browse/OSH-496

Added filter and test for error[too-many] findings in csdiff in order to apply the filter deterministically even if the location and the number of occurrences vary

This PR will should be in draft until following MR: https://github.com/csutils/csdiff/pull/185 is merged

### DIFF
--- a/src/lib/defect.hh
+++ b/src/lib/defect.hh
@@ -65,6 +65,15 @@ struct DefEvent {
         event(event)
     {
     }
+
+    void clearLoc()
+    {
+        this->fileName.clear();
+        this->line = 0;
+        this->column = 0;
+        this->hSize = 0;
+        this->vSize = 0;
+    }
 };
 
 ///< return (end - beg) if it is positive and fits into target type, 0 otherwise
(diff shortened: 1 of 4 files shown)


src/lib/msg-filter.cc
```diff
@@ -78,6 +78,8 @@ MsgFilter::MsgFilter():
     d(new Private)
 {
     d->addMsgFilter("", "[0-9][0-9]* out of [0-9][0-9]* times");
+    // ignore error[too-many] events
+    d->addMsgFilter("", "[0-9]+ occurrences of warning\\[.*\\] exceeded the specified limit [0-9]+");
```
lzaoral, line 82: While this make the diffs in tests the same, this change is wrong.  This will combine all `error[too-many]` into a single warning regardless of the actual rate-limited warning!

Please, provide the third `addMsgFilter` argument with the proper replacement and also extend the new test with such inputs.  Thank you!

lzaoral (changes requested):

PR #188, opened 2024-06-12
msg-filters: added filter for too-many event in csdiff

jperezdealgaba:
Resolves: issues.redhat.com/browse/OSH-496

Added filter and test for error[too-many] findings in csdiff in order to apply the filter deterministically even if the location and the number of occurrences vary


src/lib/msg-filter.cc
```diff
@@ -78,6 +78,8 @@ MsgFilter::MsgFilter():
     d(new Private)
 {
     d->addMsgFilter("", "[0-9][0-9]* out of [0-9][0-9]* times");
+    // ignore error[too-many] events
+    d->addMsgFilter("", "[0-9]+ occurrences of (warning\\[.*\\] exceeded the specified limit [0-9]+)", "\\1" );
```
jperezdealgaba, line 82: @lzaoral I mistakenly closed last PR. I have extended the tests and the filter is now being applied to all [tooMany] errors taking into account the rate-limit

kdudka: I think we should just replace both the numbers by a placeholder, say `NNNN`, like we do here: https://github.com/csutils/csdiff/blob/be1def43618608bbd4b662beb0053824d41042e1/src/lib/msg-filter.cc#L120

Everything else in the message should stay as it is.

I can see that some of the existing filters drop too much data, too, but we cannot change them without changing the `csdiff/v1` hashes.

jperezdealgaba: @kdudka I have replaced both numbers with a placeholder. 
I do not completely understand the changes here from the first version. If we are replacing both values by the the placeholder, wouldn't be the same for all error[too-many] findings to be removed by not providing a replacement? What am I misunderstanding?

Thanks in advance!

kdudka: Which first version?  The one proposed in #187 discarded the difference between too many `warning[cpp/UserControlledPointer]` findigns and too many `warning[cpp/IntegerOverflow]` findings, which is certainly not desired.

The values we get after the substitution are used for computation of SARIF fingerprints so we want to preserve as much relevant information as possible to avoid potential collisions with other completely unrelated findings.

kdudka (changes requested):
@jperezdealgaba Please update the regression test such that it covers the code change.  As it is now, if I revert the code change, the test is still passing.

jperezdealgaba:
@kdudka I can not reply to your comment directly. Test cases added to reflect the new changes.

kdudka (approved):
Thanks for the update!  I have anchored the regex to match only full strings and updated the commit message to describe what the change actually does.

PR #189, opened 2024-06-24
msg-filter: abstract out notes about builtin models

kdudka:
... so that semantically equal findings are matched with each other

Resolves: https://issues.redhat.com/browse/OSH-700


rhyw:
I had a quick verification, the difference are no longer shown.

LGTM

rhyw (approved):

kdudka:
@rhyw Thank you for testing it!

PR #196, opened 2024-07-23
introduce the csdiff-static subpackage

kdudka:
... with a statically linked csgrep-static executable needed for context embedding in legacy build environments.

### DIFF
--- a/make-srpm.sh
+++ b/make-srpm.sh
@@ -90,6 +90,13 @@ cat > "$SPEC" << EOF
 %bcond_without python2
 %endif
 
+# build csdiff-static on RHEL-10+ and Fedora
+%if 0%{?rhel} > 9 || 0%{?fedora}
+%bcond_without static
+%else
+%bcond_with static
+%endif
+
 # python3 support is optional
 %bcond_without python3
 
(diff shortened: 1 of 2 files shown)


kdudka:
The `csdiff-static` subpackage has been experimentally used in a Konflux CI task for source code context embedding into SAST scanning results produced by GCC Analyzer: https://github.com/ralphbean/logging-vector/pull/53/commits/529d330f18150e6fe310b95659669ff53e2fe59c

make-srpm.sh
```diff
@@ -90,6 +90,13 @@ cat > "$SPEC" << EOF
 %bcond_without python2
 %endif
 
+# build csdiff-static on RHEL-10+ and Fedora
+%if 0%{?rhel} > 9 || 0%{?fedora}
```
siteshwar, line 94: The public instance of OpenScanHub uses `c9s` for workers. So, this should not affect existing deployment.

siteshwar (approved):
No blockers from my side.

make-srpm.sh
kdudka: I believe that nothing changes for existing users on Fedora either.  The `csdiff-static` subpackage will not be installed unless it is explicitly requested.  And, even if it was installed, the `csgrep-static` binary will not appear in default search paths.  So it cannot be accidentally used.

kdudka:
@siteshwar Thanks for review!

Issue #193, opened 2024-07-23
Shall we use other term than "defects" in the reports?


siteshwar:
This is a follow up on [fedora-devel comment](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/PXTDMU3K3QTGZKOIJ2PRR6QRID7LQRVV/).

Not all code paths identified to contain possible bugs may be real bugs. There is objection on the use of term "defects" and "flaws" for such code paths. This issue tracks using more appropriate term for such code paths.

kdudka:
The Product Security team at Red Hat tends to call them **findings**.

@siteshwar @lzaoral Would it make sense to use this term upstream as well?

lzaoral:
> Would it make sense to use this term upstream as well?

Sure, why not.  I don't have a strong opinion about this.

berrange:
FYI, from my pov as a maintainer, I dislike any terminology that suggests the reported items are flaws in my projects. Given the well known high false positive rates, it is defamatory to the reputation of the project to suggest they are bugs, defects or flaws, until a human has weeded out the false positives.

"Findings" is a reasonable term that doesn't pre-suppose whether the identified items are bugs or not.

PR #197, opened 2024-07-25
html-writer: use the term "finding"

siteshwar:
... instead of "defect" in the reports.

Resolves: https://github.com/csutils/csdiff/issues/193


kdudka (commented):
Looks good.  I would also rename the `newDefect` CSS class to `newFinding` for consistency.

siteshwar:
There is also the use of `#def` in the reports. Shall that be fixed too?

kdudka:
Yep, I would keep `#def` to avoid unnecessary changes in URLs to scan results.

kdudka (approved):

PR #205, opened 2024-09-09
parser-json-sarif: initialize the `imp` flag

kdudka:
... to 1 if `level` is `error`.  This is how `csgrep ---mode=sarif` encodes the `imp` flag.  As a side effect, all findings from Snyk Code with `level` set to `error` will be marked as important.  Nevertheless, this is what `csmock-plugin-snyk` explicitly does already: https://github.com/csutils/csmock/pull/122

Resolves: https://issues.redhat.com/browse/OSH-754
Closes: https://github.com/csutils/csdiff/pull/205


lzaoral (approved):

kdudka:
@lzaoral Thanks for review!

PR #211, opened 2024-10-07
writer-html: reword warnings about Coverity parsing

kdudka:
... to make them more self-explaining

Resolves: https://issues.redhat.com/browse/OSH-422


jamacku (approved):

kdudka:
@jamacku Thanks for review!

PR #212, opened 2024-10-15
csgrep --file-glob: expand glob patterns

kdudka:
... in the names of input files.  If we have too many input files, we cannot rely on the glob expansion implemented in the shell because the expanded command might exceed the maximum command line length.  Running `csgrep` for multiple batches of input files is difficult when we use a JSON-based output format, which cannot be concatenated as plain text.

Related: https://github.com/csutils/csmock/pull/187


lzaoral (approved):
Note that I did not actually review the changes since I do not work this project any more.

PR #215, opened 2024-11-08
csdiff-core: include correct header file

kdudka:
This module does not use any specific writer.  It rather uses the generic `createWriter()` function, which is declared in `writer.hh`.


jamacku (approved):
LGTM

kdudka:
@jamacku Thanks for review!

PR #217, opened 2025-02-05
cstrans-df-run: introduce the --shell-form option

kdudka:
... to write transformed RUN lines using the shell form (rather than exec form, which is the default format).

Related: https://issues.redhat.com/browse/KFLUXBUGS-1943
Related: https://issues.redhat.com/browse/OSH-809

PR #219, opened 2025-02-07
tests/cstrans-df-run: fix the macro that adds a test case

kdudka:
... so that tests with/without the `--shell-form` option get assigned distinct test names rather than effectively overwriting each other. This one-line change enables additional 11 tests.

Fixes: commit b05206d17d6a7a3fadd689c6674f805344aafd00
Related: https://github.com/csutils/csdiff/pull/217
Closes: https://github.com/csutils/csdiff/pull/219


jamacku (approved):
LGTM

kdudka:
@jamacku Thanks for review!

PR #220, opened 2025-02-17
ci: upload SARIF file only when debug is enabled

jamacku:


kdudka (approved):
Seems to work as expected.  Thanks!

PR #224, opened 2025-04-07
msg-filter: tweak path substitution for RHIVOS kernels

kdudka:
... that use the `el9_6iv` dist tag.  The regex we have been using did not accept the `iv` suffix.

Reported-by: Erico Nunes

### DIFF
--- a/src/lib/msg-filter.cc
+++ b/src/lib/msg-filter.cc
@@ -298,7 +298,7 @@ std::string MsgFilter::filterPath(
 
     // try to kill the multiple version strings in paths (kernel, OpenLDAP, ...)
     static const std::string strKrn = "^[a-zA-Z+]+";
-    static const RE reKrn(strKrn + /* convert el8_9 -> el8 */ "|_[0-9]+$");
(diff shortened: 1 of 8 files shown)


sfowl (approved):
LGTM.

FWIW, looking at some internal data, suffixes like "el8_9.1" also seem somewhat common. Here's a few examples:

```
 autofs-5.1.4-109.el8_9.1
 avahi-0.7-21.el8_9.1
 bind-9.11.36-11.el8_9.1
 bind9.16-9.16.23-0.16.el8_9.2
 c-ares-1.13.0-9.el8_9.1
 ```

kdudka:
Yes, this is the recommended versioning of Z-stream updates.  These suffixes do not appear in file paths in scan results though.  For example, if you look at the scan results of `curl-8.12.1-1.el10` and `curl-8.12.1-1.el10_0.1`, the top-level directory is `curl-8.12.1` in both the cases.  So these results can be easily matched with each other without any transformation.  The regex we are patching in this pull requst is mostly used for Z-stream **kernel** builds, which use a very specific directory structure, as described in the commit message of f05aa13b32404376f9c673604205c37693715469.

PR #228, opened 2025-04-30
parser-json-sarif: canonicalize paths from GCC SARIF

kdudka:
... so that we do not end up with paths like this:
```
usbutils-018/redhat-linux-build/../usbmisc.c
```

Related: https://github.com/openscanhub/known-false-positives/pull/6


siteshwar (approved):

kdudka:
@siteshwar Thanks for review!

